PR #96, opened 2022-10-11
Factor out UI custom bottloader logic

italo-sampaio:
Moved custom bootloader logic to specific modules apart from custom UI code.


italo-sampaio:
This will be addressed in other PRs

PR #97, opened 2022-10-24
Factor out UI modules unlock and pin

italo-sampaio:
- Factor out unlock and pin modules
- Add unit tests for unlock and pin

### DIFF
--- a/ledger/src/ui/src/bolos_ux.c
+++ b/ledger/src/ui/src/bolos_ux.c
@@ -52,16 +52,13 @@
 #include "attestation.h"
 #include "signer_authorization.h"
 #include "memutil.h"
+#include "unlock.h"
 
 // Onboarded with the UI flag
 const unsigned char N_onboarded_ui[1] = {0};
 
 // PIN buffer used for authenticated operations
-unsigned char G_pin_buffer[PIN_LENGTH + 2];
-// Skip the prepended length of pin buffer
-#define G_PIN_BUFFER_PAYLOAD (G_pin_buffer + 1)
-// Unify pin buffer length
-#define G_PIN_BUFFER_LEN() strlen((const char *)G_PIN_BUFFER_PAYLOAD)
+unsigned char G_pin_buffer[MAX_PIN_LENGTH + 2];
 
 #ifdef OS_IO_SEPROXYHAL
 
@@ -243,6 +240,9 @@ void io_seproxyhal_display(const bagl_element_t *element) {
 // Signer authorization context shorthand
 #define sigaut_ctx (G_bolos_ux_context.sigaut)
 
+// Pin context shorthand
+#define pin_ctx (G_bolos_ux_context.pin)
+
 // Operation being currently executed
 static unsigned char curr_cmd;
 
@@ -265,6 +265,7 @@ static void reset_if_starting(unsigned char cmd) {
         G_bolos_ux_context.words_buffer_length = 0;
         reset_attestation(&attestation_ctx);
         reset_signer_authorization(&sigaut_ctx);
+        reset_pin(&pin_ctx);
     }
 }
 
@@ -321,11 +322,9 @@ static void sample_main(void) {
                     break;
                 case RSK_PIN_CMD: // Send pin_buffer
                     reset_if_starting(RSK_META_CMD_UIOP);
-                    pin = APDU_AT(2);
-                    if ((pin >= 0) && (pin <= PIN_LENGTH)) {
-                        G_pin_buffer[pin] = APDU_AT(3);
-                        G_pin_buffer[pin + 1] = 0;
-                    }
+                    tx = pin_cmd(&pin_ctx);
+                    // Save pin for authenticated operations
+                    get_pin_ctx(&pin_ctx, G_pin_buffer);
                     THROW(APDU_OK);
                     break;
                 case RSK_IS_ONBOARD: // Wheter it's onboarded or not
@@ -347,8 +346,9 @@ static void sample_main(void) {
                     nvm_write(
                         (void *)PIC(N_onboarded_ui), (void *)&aux, sizeof(aux));
 
+                    set_pin_ctx(&pin_ctx, G_pin_buffer);
 #ifndef DEBUG_BUILD
-                    if (!is_pin_valid(G_PIN_BUFFER_PAYLOAD)) {
+                    if (!is_pin_valid(&pin_ctx)) {
                         THROW(ERR_INVALID_PIN);
                     }
 #endif
@@ -392,14 +392,18 @@ static void sample_main(void) {
                                    sizeof(G_bolos_ux_context.words_buffer));
                     // Set PIN
                     os_perso_set_pin(
-                        0, G_PIN_BUFFER_PAYLOAD, G_PIN_BUFFER_LEN());
+                        0,
+                        pin_ctx.pin_buffer.payload,
+                        strlen((const char *)pin_ctx.pin_buffer.payload));
                     // Finalize onboarding
                     os_perso_finalize();
                     os_global_pin_invalidate();
                     SET_APDU_AT(1, 2);
-                    SET_APDU_AT(2,
-                                os_global_pin_check(G_PIN_BUFFER_PAYLOAD,
-                                                    G_PIN_BUFFER_LEN()));
+                    SET_APDU_AT(
+                        2,
+                        os_global_pin_check(
+                            pin_ctx.pin_buffer.payload,
+                            strlen((const char *)pin_ctx.pin_buffer.payload)));
                     // Clear pin buffer
                     explicit_bzero(G_pin_buffer, sizeof(G_pin_buffer));
                     // Turn the onboarding flag on to mark onboarding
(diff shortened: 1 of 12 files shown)


ledger/src/ui/src/pin.h
```diff
@@ -27,8 +27,79 @@
 
 #include <stdbool.h>
 
-#define PIN_LENGTH 8
+#define MAX_PIN_LENGTH 8
+#define MIN_PIN_LENGTH 4
```
amendelzon, line 31: Are we using `MIN` and `MAX`? I thought we were only accepting 8 character pins.

ledger/src/ui/test/pin/test_pin.c
```diff
@@ -29,48 +29,83 @@
...
     printf("Test OK...\n");
+    pin_t pin_ctx;
+    reset_pin(&pin_ctx);
 
-    assert(is_pin_valid((unsigned char*)"abcdefgh"));
-    assert(is_pin_valid((unsigned char*)"8b23ef1s"));
-    assert(is_pin_valid((unsigned char*)"MN22P3S9"));
-    assert(is_pin_valid((unsigned char*)"MN22P3S9"));
+    set_payload(&pin_ctx, (unsigned char *)"abcdefgh", strlen("abcdefgh"));
+    assert(is_pin_valid(&pin_ctx));
```
amendelzon, line 42: Suggestion: replace `set_payload()` + `assert(is_pin_valid())` calls with a call to an `assert_pin` that does both. Then tests could be more readable. Another one for the negative case (or the same with another parameter). You could also use a macro for `true` and `false`, so maybe tests cases could look like:

```
assert_pin("abcdefgh", IS_VALID);
assert_pin("12345678", IS_INVALID);
```

ledger/src/ui/test/mock/os.h
```diff
@@ -0,0 +1,53 @@
...
+                                 unsigned char pin_length);
+void os_perso_set_pin(unsigned int identity,
+                      unsigned char *pin,
+                      unsigned int length);
+void os_global_pin_invalidate(void);
+
+/**
+ * Other mocks
+ */
+void explicit_bzero(void *s, size_t len);
```
amendelzon, line 52: I think you can find `explicit_bzero` in `string.h`?

amendelzon (changes requested):
Implementation looking good so far! I've added a few comments with things I think we can improve. The one we probably need to address for sure is the pin buffer duplication. Let me know your thoughts!

ledger/src/ui/src/pin.c
```diff
@@ -33,29 +35,66 @@
...
     return hasAlpha;
+}
+
+void get_pin_ctx(pin_t* pin_ctx, unsigned char* pin_buffer) {
+    memcpy(pin_buffer, pin_ctx->pin_raw, sizeof(pin_ctx->pin_raw));
+}
+
+void set_pin_ctx(pin_t* pin_ctx, unsigned char* pin_buffer) {
+    memcpy(pin_ctx->pin_raw, pin_buffer, sizeof(pin_ctx->pin_raw));
+}
```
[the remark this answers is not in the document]
italo-sampaio: Yeah, there isn't an easy solution to this. We need a copy of the pin outside `G_bolos_ux_context` since the context is reset on every new command. Your approach works, I'll change to that.

ledger/src/ui/src/pin.h
```diff
@@ -27,8 +27,79 @@
...
 
-bool is_pin_valid(unsigned char *pin);
+// Pin context
+typedef struct {
+    union {
+        struct {
+            unsigned char length;
+            unsigned char payload[MAX_PIN_LENGTH + 1];
+        } pin_buffer;
+        unsigned char pin_raw[MAX_PIN_LENGTH + 2];
```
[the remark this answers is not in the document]
italo-sampaio: The tricky part here is that some commands fill the buffer with the prepended length, while others don't. So the manipulation of the buffer is different depending on the command. I will change pin_t to just hold a pointer and add helper functions to manipulate that, as you suggested.

ledger/src/ui/src/pin.h
italo-sampaio: We accept pins with less than 8 chars on debug mode, but you are correct, we never use the `MIN_PIN_LENGTH` macro. I'll change that to `PIN_LENGTH`.

ledger/src/ui/src/unlock.h
```diff
@@ -0,0 +1,40 @@
...
+
+/*
+ * Implements RSK UNLOCK command.
+ *
+ * Unlocks the device.
+ *
+ * @arg[in] pin_ctx pin context
+ * @ret             number of transmited bytes to the host
+ */
+unsigned int unlock(pin_t *pin_ctx);
```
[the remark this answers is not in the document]
italo-sampaio: TBH I'm not a big fan of maintaining duplicated comments on both files, but I agree that the most important thing is that the approach is consistent. I'll follow the attestation approach and keep both comments.

ledger/src/ui/test/pin/test_pin.c
italo-sampaio: That looks great, it will improve readability a lot, I'll change that.

ledger/src/ui/test/mock/os.h
italo-sampaio: My bad, I'll change that =)

ledger/src/ui/src/unlock.h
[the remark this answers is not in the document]
amendelzon: I'm happy to change to a new comment style, not a fan of duplicated comments either. Let's agree on one and start implementing it after this PR.

ledger/src/ui/test/mock/os.h
amendelzon: No worries, thanks :)

ledger/src/ui/src/bolos_ux.c
```diff
@@ -321,11 +322,8 @@ static void sample_main(void) {
                     break;
                 case RSK_PIN_CMD: // Send pin_buffer
                     reset_if_starting(RSK_META_CMD_UIOP);
-                    pin = APDU_AT(2);
-                    if ((pin >= 0) && (pin <= PIN_LENGTH)) {
-                        G_pin_buffer[pin] = APDU_AT(3);
-                        G_pin_buffer[pin + 1] = 0;
-                    }
+                    init_pin_ctx(&pin_ctx, G_pin_buffer);
+                    tx = update_pin_buffer(&pin_ctx);
```
amendelzon, line 326: Just noticing this now: we're probably missing `rx` as a parameter, and corresponding validations (which are currently non-existent, but we should probably add).

ledger/src/ui/src/bolos_ux.h
```diff
@@ -160,13 +160,11 @@ typedef struct bolos_ux_context {
...
             sigaut_t sigaut;
+            pin_t pin;
         };
     };
 
-#define MAX_PIN_LENGTH 8
-#define MIN_PIN_LENGTH 4
-    char pin_buffer[MAX_PIN_LENGTH +
-                    1]; // length prepended for custom pin length
+    char pin_buffer[PIN_LENGTH + 1]; // length prepended for custom pin length
```
amendelzon, line 167: do we need this here at all?

ledger/src/ui/src/pin.h
```diff
@@ -29,6 +29,67 @@
...
 #define PIN_LENGTH 8
 
-bool is_pin_valid(unsigned char *pin);
+// Pin context
+typedef struct {
+    unsigned char* pin_buffer;
+} pin_t;
+
+// Helper macros for pin context manipulation
+#define PIN_CTX_PAYLOAD(pin_ctx) ((unsigned char*)((pin_ctx)->pin_buffer + 1))
```
amendelzon, line 38: Two things that are not super relevant, but might aid in readability:

1. I feel like we need a high level explanation of the two pin versions we use, and where each is used (i.e., with prepended length and null-terminated).
2. The names `PIN_CTX_PAYLOAD` and `PIN_CTX_PAYLOAD_LEN` might be a bit over complicated. Maybe something simpler and more to the point? Like `GET_PIN` and `GET_PIN_LENGTH`? Or something along those lines?

What do you think?

ledger/src/ui/test/mock/os.h
amendelzon: Taking this back, apparently in the `string.h` version we are using to build these, `explicit_bzero` is not available. Also take a look at the TCPSigner implementation of this function, it has got a compiler barrier to prevent it from optimizing such calls.

ledger/src/ui/test/pin/test_pin.c
```diff
@@ -29,52 +29,73 @@
 
 #include "pin.h"
 
+#define IS_VALID true
+#define NOT_VALID false
```
amendelzon, line 33: `IS_NOT_VALID`?

amendelzon (changes requested):
LGTM. Apart from a variable that we probably don't need, very minor suggestions. Please take a look and let me know your thoughts. Thanks!

amendelzon (approved):
Looks good to me, thanks for tackling this!

PR #98, opened 2022-11-01
Factored out onboard and communication UI modules

italo-sampaio:
- Factored out onboard-related operations
- Factored out communication module
- Incidentally removed unused `rx` parameter from unlock and pin modules


ledger/src/ui/src/bolos_ux.c
```diff
@@ -326,86 +320,15 @@ static void sample_main(void) {
...
-                    os_perso_set_pin(
-                        0, GET_PIN(&pin_ctx), GET_PIN_LENGTH(&pin_ctx));
-                    // Finalize onboarding
-                    os_perso_finalize();
-                    os_global_pin_invalidate();
-                    SET_APDU_AT(1, 2);
-                    SET_APDU_AT(2,
-                                os_global_pin_check(GET_PIN(&pin_ctx),
-                                                    GET_PIN_LENGTH(&pin_ctx)));
+                    tx = wipe_device(&onboard_ctx, &pin_ctx);
```
amendelzon, line 329: This should probably be something more along the lines of `onboard_device` since we not only wipe it, but mainly onboard it.

ledger/src/ui/src/communication.c
```diff
@@ -0,0 +1,65 @@
...
+ */
+unsigned int echo(unsigned int rx) {
+    return rx;
+}
+
+/*
+ * Implement the RSK MODE command.
+ *
+ * Since the UI only runs on BOOTLOADER mode, this always returns
+ * RSK_MODE_BOOTLOADER
```
amendelzon, line 44: I'd say that "Since this is only ever called from the bootloader,..."

ledger/src/ui/src/onboard.c
```diff
@@ -0,0 +1,160 @@
...
+#include <string.h>
+
+#include "apdu.h"
+#include "defs.h"
+#include "err.h"
+#include "os.h"
+#include "onboard.h"
+
+// Global onboarding flag
+extern const unsigned char *N_onboarded_ui[1];
```
amendelzon, line 34: SInce this is an onboarding-related variable, I would probably move this definition here and declare as `extern` elsewhere when used.

ledger/src/ui/src/onboard.c
```diff
@@ -0,0 +1,160 @@
...
+ * @ret                  number of transmited bytes to the host
+ */
+unsigned int set_client_seed(volatile unsigned int rx, onboard_t *onboard_ctx) {
+    // Should receive 1 byte per call
+    if (APDU_DATA_SIZE(rx) != 1) {
+        THROW(PROT_INVALID);
+    }
+
+    unsigned char index = APDU_OP();
+    if ((index >= 0) && ((size_t)index <= sizeof(onboard_ctx->words_buffer))) {
```
amendelzon, line 138: I think this should fill `client_seed` and the `cx_rng` should fill another buffer called `seed`, that we then XOR into `seed`. That is then used to generate `words_buffer`. So then maybe `words_buffer` and `client_seed` can go into a `union`. Makes sense? Also we might find use in rewording "client" to "host". Up to you :)

ledger/src/ui/src/onboard.c
```diff
@@ -0,0 +1,160 @@
...
+    // Set PIN
+    os_perso_set_pin(0, GET_PIN(pin_ctx), GET_PIN_LENGTH(pin_ctx));
+    // Finalize onboarding
+    os_perso_finalize();
+    os_global_pin_invalidate();
+    unsigned char output_index = CMDPOS;
+    SET_APDU_AT(output_index++, 2);
+    SET_APDU_AT(output_index++,
+                os_global_pin_check(GET_PIN(pin_ctx), GET_PIN_LENGTH(pin_ctx)));
+
```
amendelzon, line 110: I think we need to clear the pin after validating it.

ledger/src/ui/src/onboard.c
amendelzon, line 138: Also the client or host seed should be 32 bytes long.

ledger/src/ui/src/onboard.h
```diff
@@ -0,0 +1,85 @@
...
+#include "os.h"
+#include "os_io_seproxyhal.h"
+#include "pin.h"
+
+// passphrase will be prefixed with "MNEMONIC" from BIP39, the passphrase
+// content shall start @ 8
+unsigned int bolos_ux_mnemonic_from_data(unsigned char *in,
+                                         unsigned int inLength,
+                                         unsigned char *out,
+                                         unsigned int outLength);
```
amendelzon, line 37: I think we should probably create a header file for `bolos_ux_onboarding_seed_bip39.c` and include that rather than having that header casually lying around here. What do you think?

ledger/src/ui/src/onboard.h
```diff
@@ -0,0 +1,85 @@
...
+
+// Onboard context
+typedef struct {
+    char client_seed[MAX(
```
amendelzon, line 41: The client (or host) seed will always be 32 bytes. We don't need the `MAX` cause now we're not using this buffer for anything else.

ledger/src/ui/src/onboard.h
```diff
@@ -0,0 +1,85 @@
...
+        32, sizeof(bagl_icon_details_t) + BOLOS_APP_ICON_SIZE_B - 1)];
+    char words_buffer[257];
```
amendelzon, line 43: Does it make sense to declare this as `unsigned char *`?

amendelzon (changes requested):
Looking great so far! I think there's a few things we need to tackle before we can think of getting this merged in, but we're not far off!

ledger/src/ui/src/onboard.h
```diff
@@ -29,18 +29,13 @@
...
-                                         unsigned char *out,
-                                         unsigned int outLength);
-
 // Onboard context
 typedef struct {
-    char client_seed[MAX(
-        32, sizeof(bagl_icon_details_t) + BOLOS_APP_ICON_SIZE_B - 1)];
-    char words_buffer[257];
+    union {
+        unsigned char words_buffer[257];
```
amendelzon, line 35: Is there a constant available we can use instead of plain `257`? Maybe within the `bolos_ux_mnemonic_from_data` file there is or we could add one instead?

ledger/src/ui/src/onboard.c
```diff
@@ -70,29 +71,30 @@ unsigned int wipe_device(onboard_t *onboard_ctx, pin_t *pin_ctx) {
     os_global_pin_invalidate();
     os_perso_wipe();
     // Generate 32 bytes of random with onboard rng
-    cx_rng((unsigned char *)onboard_ctx->client_seed, HASHSIZE);
+    cx_rng((unsigned char *)onboard_ctx->seed, sizeof(onboard_ctx->seed));
     // XOR with host-generated 32 bytes random
-    for (int i = 0; i < HASHSIZE; i++) {
-        onboard_ctx->client_seed[i] ^= onboard_ctx->words_buffer[i];
+    for (int i = 0; i < SEEDSIZE; i++) {
```
amendelzon, line 76: I think it's safer to use `sizeof()` here instead of `SEEDSIZE`.

ledger/src/ui/src/onboard.c
```diff
@@ -70,29 +71,30 @@ unsigned int wipe_device(onboard_t *onboard_ctx, pin_t *pin_ctx) {
...
+        onboard_ctx->seed[i] ^= onboard_ctx->host_seed[i];
     }
-    // The seed is now in client_seed, generate the mnemonic
+    // The seed is now in onboard_ctx->seed, generate the mnemonic
     os_memset(onboard_ctx->words_buffer, 0, sizeof(onboard_ctx->words_buffer));
     onboard_ctx->words_buffer_length =
-        bolos_ux_mnemonic_from_data((unsigned char *)onboard_ctx->client_seed,
+        bolos_ux_mnemonic_from_data((unsigned char *)onboard_ctx->seed,
                                     SEEDSIZE,
```
amendelzon, line 83: Same here re `sizeof()`.

amendelzon (changes requested):
Almost there! There's a few minor changes we need to do re safety and we're practically ready for unit tests!

ledger/src/ui/src/onboard.c
amendelzon: I just saw that now, sorry. And I also saw that we also clear the pin manually on every operation in that way. I think it's fine for now, but we should probably turn that into a `clear_pin`/`zero_pin` function eventually and call it within every module as part of its functionality. Thoughts?

amendelzon: We'll just keep the clearing of the pin within the calling code since, as @italo-sampaio mentioned, it is indeed clearer to see which operations preseve the pin and which don't in that way.

ledger/src/ui/src/onboard.c
```diff
@@ -73,16 +72,15 @@ unsigned int onboard_device(onboard_t *onboard_ctx, pin_t *pin_ctx) {
     // Generate 32 bytes of random with onboard rng
     cx_rng((unsigned char *)onboard_ctx->seed, sizeof(onboard_ctx->seed));
     // XOR with host-generated 32 bytes random
-    for (int i = 0; i < SEEDSIZE; i++) {
+    for (unsigned int i = 0; i < sizeof(onboard_ctx->seed); i++) {
         onboard_ctx->seed[i] ^= onboard_ctx->host_seed[i];
     }
     // The seed is now in onboard_ctx->seed, generate the mnemonic
     os_memset(onboard_ctx->words_buffer, 0, sizeof(onboard_ctx->words_buffer));
```
amendelzon, line 79: I would use `explicit_bzero` here since we really want to make sure that buffer is zeroed out.

ledger/src/ui/src/pin.c
```diff
@@ -108,24 +70,79 @@ unsigned int update_pin_buffer(volatile unsigned int rx, pin_t* pin_ctx) {
...
+
+/*
+ * Fills the internal pin buffer with zeroes
+ */
+void clear_pin() {
+    explicit_bzero(G_pin_buffer, sizeof(G_pin_buffer));
+}
+
+/*
+ * Validates the pin currently saved to the internal pin buffer
```
amendelzon, line 127: I would try to be more explicit with the naming and description here so as to not to confuse this with pin rules validation. This method actually unlocks the device, giving access to secrets usage, so I'd try and find a name closer to that description.

amendelzon (dismissed):
This looks great! Really liked the pin module refactor, things are more readable now IMHO. Thanks again for tackling this!

ledger/src/ui/src/onboard.c
```diff
@@ -0,0 +1,160 @@
...
+
+#include "apdu.h"
+#include "bolos_ux_onboarding_seed_bip39.h"
+#include "defs.h"
+#include "err.h"
+#include "os.h"
+#include "onboard.h"
+
+// Global onboarding flag
+const unsigned char *N_onboarded_ui[1];
```
amendelzon, line 35:
```suggestion
const unsigned char N_onboarded_ui[1];
```

ledger/src/ui/src/onboard.c
```diff
@@ -0,0 +1,160 @@
...
+ * @ret                  number of transmited bytes to the host
+ */
+unsigned int set_host_seed(volatile unsigned int rx, onboard_t *onboard_ctx) {
+    // Should receive 1 byte per call
+    if (APDU_DATA_SIZE(rx) != 1) {
+        THROW(PROT_INVALID);
+    }
+
+    unsigned char index = APDU_OP();
+    if ((index >= 0) && ((size_t)index <= sizeof(onboard_ctx->host_seed))) {
```
amendelzon, line 138:
```suggestion
    if ((index >= 0) && ((size_t)index < sizeof(onboard_ctx->host_seed))) {
```

ledger/src/ui/src/onboard.h
```diff
@@ -0,0 +1,81 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#ifndef __ONBOARD
+#define __ONBOARD
+
+#include "os.h"
+#include "os_io_seproxyhal.h"
+#include "pin.h"
```
amendelzon, line 30: Do we need all of these here?

amendelzon (changes requested):
Minor (but some important) additional comments.

ledger/src/ui/src/onboard.h
italo-sampaio: nice catch, we don't!

amendelzon (approved):
Looking great! Ready for unit tests!

ledger/src/ui/test/communication/test_communication.c
```diff
@@ -0,0 +1,59 @@
...
+    printf("Test get mode...\n");
+    assert(2 == get_mode());
+    assert(RSK_MODE_BOOTLOADER == APDU_AT(1));
+}
+
+void test_get_retries() {
+    printf("Test get retries...\n");
+
+    assert(3 == get_retries());
+    assert(MOCK_INTERNAL_RETRIES_COUNTER == APDU_AT(2));
```
amendelzon, line 50: Instead of a constant, I'd add a mocking function that sets the mocked value and then check against that. In that way it's a bit more flexible and the test is self contained.

amendelzon (approved):
Looks good, retested and everything works properly on the physical device. Thanks for tackling this one!

PR #103, opened 2022-11-29
Factored out UI bootloader module

italo-sampaio:
- Factored out bootloader main function
- Factored out handlers for bolos ux ids
- Added "bootloader mode" to `bootloader_main()` which doesn't allow commands after onboard


ledger/src/ui/src/bootloader.c
```diff
@@ -0,0 +1,373 @@
...
+    if (cmd != curr_cmd) {
+        curr_cmd = cmd;
+        reset_attestation(&attestation_ctx);
+        reset_signer_authorization(&sigaut_ctx);
+        reset_onboard_ctx(&onboard_ctx);
+    }
+}
+
+// Check if we allow this version of the app to execute.
+static int is_app_version_allowed(application_t *app) {
```
amendelzon, line 67: Looking now at the uses of this function, it probably doesn't make a lot of sense to have it. What do you think?

ledger/src/ui/src/bootloader.c
```diff
@@ -0,0 +1,373 @@
...
+                    tx = unlock();
+                    // The pin value will also be used in
+                    // BOLOS_UX_CONSENT_APP_ADD command, so we can't wipe the
+                    // pin buffer here
+                    THROW(APDU_OK);
+                    break;
+                case RSK_END_CMD: // return to dashboard
+                    reset_if_starting(RSK_END_CMD);
+                    if (onboarding_mode) {
+                        while (1)
```
amendelzon, line 197: Maybe we misunderstood each other when planning this. I thought the idea was to disallow any additional operations from executing after the `RSK_WIPE` command ended. This only hangs the device when the `RSK_END_CMD` command is executed, but nothing prevents a user from executing other commands instead.

ledger/src/ui/src/bootloader.c
```diff
@@ -0,0 +1,373 @@
...
+                            // the onboard, so we never return from
+                            // bootloader_main()
+                            ;
+                    }
+                    autoexec = 1;
+                    return;
+                case RSK_END_CMD_NOSIG: // return to dashboard
+                    reset_if_starting(RSK_END_CMD_NOSIG);
+                    if (onboarding_mode) {
+                        while (1)
```
amendelzon, line 208: Same here.

ledger/src/ui/src/bootloader.c
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    io_seproxyhal_init();
+    USB_power(1);
+    screen_settings_apply();
+    screen_not_personalized_init();
+    bootloader_main(true);
+    // bootloader_main() actually never returns when onboarding mode is active,
+    // so this value is never actually returned to the caller
+    return BOLOS_UX_OK;
```
amendelzon, line 284: We should probably return `BOLOS_UX_CANCEL` here instead.

ledger/src/ui/src/bootloader.c
```diff
@@ -0,0 +1,373 @@
...
+            }
+        }
+        i++;
+    }
+}
+
+// Main function for the bootloader. If onboarding_mode == true, this function
+// never returns after RSK_END_CMD, so the user is required to unplug the
+// device before proceeding
+static void bootloader_main(bool onboarding_mode) {
```
amendelzon, line 94: Looking at the `onboarding_mode` and its usages, it might be a bit more declarative to use an enum `mode` with two defined values instead. Just thinking about readability. What do you think?

amendelzon (changes requested):
Looking great so far. A few comments regarding some things I think we need to tackle and then onto unit testing.

ledger/src/ui/src/bolos_ux.c
```diff
@@ -483,77 +278,34 @@ void bolos_ux_main(void) {
...
-            io_seproxyhal_setup_ticker(100);
-            // Run first application once
 
-            if (autoexec) {
-                autoexec = 0;
-                run_first_app();
-            }
-            screen_dashboard_init();
+        case BOLOS_UX_DASHBOARD:
+            handle_bolos_ux_boot_dashboard();
```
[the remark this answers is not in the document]
italo-sampaio: My understanding is that all the calls that show a new screen, like this one, set the field `exit_code_after_elements_displayed` on `G_bolos_ux_context`. This field is used in other places in `bolos_ux.c` to set the actual value of `exit_code`.  With that in mind, it felt misleading/redundant to set this field here, since the actual value returned is handled by the screen logic. What do you think?

ledger/src/ui/src/bootloader.c
italo-sampaio: Hmmm, my bad here. I will change this to hang after `RSK_WIPE`

ledger/src/ui/src/bootloader.c
italo-sampaio: Agreed, that should be clearer

ledger/src/ui/src/bootloader.c
italo-sampaio: I hadn't noticed that before, but since you mentioned maybe it makes sense to get rid of this function and use `is_authorized_signer` directly, what do you think?

ledger/src/ui/src/bolos_ux.c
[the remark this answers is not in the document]
amendelzon: Good point, I had no clue. Let's just leave it as is then.

ledger/src/ui/src/bootloader.c
amendelzon: I agree, let's do that.

ledger/src/ui/src/bootloader.c
amendelzon: I would try to avoid hanging/leaving the USB unresponsive. Instead, after onboarding I'd just raise an error on any command. How does that sound?

amendelzon (dismissed):
Looks good to me, ready for tests!

ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -0,0 +1,536 @@
...
+
+    // Run bootloader_main and send RSK_END_CMD to set autoexec = 1
+    pthread_mutex_lock(&mutex);
+    reset_mock_apdu_buffer();
+    add_mock_apdu_buffer(RSK_END_CMD);
+
+    pthread_t onboarding_thread;
+    pthread_create(&onboarding_thread, NULL, bootloader_main_wrapper, NULL);
+
+    // RSK_END_CMD_NOSIG returns nothing, we just need to release io_exchange()
```
italo-sampaio, line 413: typo: RSK_END_CMD

ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -0,0 +1,536 @@
...
```
italo-sampaio, line 382: typo: RSK_END_CMD

italo-sampaio (commented):
Some typos on comments, will fix those post-review

ledger/src/ui/test/bootloader/mock/mock.h
```diff
@@ -0,0 +1,96 @@
...
+
+// Empty struct used to mock data types
+struct mock_struct {
+    void* mock_data;
+};
+/*
+typedef struct mock_application_s {
+    unsigned int flags;
+    unsigned char hash[32];
+} application_t;
```
italo-sampaio, line 58: Dead code: I'll remove this

ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -0,0 +1,619 @@
...
+
+void reset_signer_authorization(sigaut_t* sigaut_ctx) {
+    G_reset_signer_authorization_called = true;
+}
+
+void reset_onboard_ctx(onboard_t* onboard_ctx) {
+    G_reset_onboard_called = true;
+}
+
+// Function definitions required for compiling bootloade.c
```
italo-sampaio, line 180: typo: bootloader.c

ledger/src/ui/test/bootloader/mock/mock.h
```diff
@@ -0,0 +1,96 @@
...
+
+#define IO_APDU_BUFFER_SIZE (5 + 255)
+extern unsigned char G_io_apdu_buffer[IO_APDU_BUFFER_SIZE];
+
+// Assert helpers
+#define ASSERT_EQUALS(a, b) assert((a) == (b))
+#define ASSERT_TRUE(cond) assert(cond)
+#define ASSERT_FALSE(cond) assert(!(cond))
+#define ASSERT_NOT_NULL(obj) assert(NULL != obj)
+#define ASSERT_APTU_AT(i, val) assert(APDU_AT(i) == val)
```
italo-sampaio, line 44: Typo on this define: should be ASSERT_APDU_AT

ledger/src/ui/src/bolos_ux_handlers.c
```diff
@@ -0,0 +1,171 @@
...
+    // apply settings when redisplaying dashboard
+    screen_settings_apply();
+
+    // when returning from application, the ticker could have been
+    // disabled
+    io_seproxyhal_setup_ticker(100);
+    // Run signer application once
+    if (autoexec) {
+        autoexec = 0;
+        run_signer_app();
```
amendelzon, line 97: Shall we also add a `return` after this to prevent the following line from running?

ledger/src/ui/src/bolos_ux_handlers.c
```diff
@@ -0,0 +1,171 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+#include "bolos_ux_common.h"
+#include "bootloader.h"
+
+volatile unsigned char autoexec = 0;
```
amendelzon, line 27: This can probably also be `static`. Also, do we need the `volatile` here?

ledger/src/ui/src/bootloader.c
```diff
@@ -0,0 +1,227 @@
...
+
+    return tx;
+}
+
+// Main function for the bootloader. If mode == BOOTLOADER_MODE_ONBOARD,
+// commands are not accepted after the onboard is performed.
+void bootloader_main(bootloader_mode_t mode) {
+    volatile unsigned int rx = 0;
+    volatile unsigned int tx = 0;
+    volatile unsigned char onboard_performed = 0;
```
amendelzon, line 191: `onboard_performed` is only used within this module, do you think a `static` global variable (such as e.g. `current_cmd`) should suffice?

ledger/src/ui/src/bootloader.c
```diff
@@ -0,0 +1,227 @@
...
+        reset_if_starting(RSK_IS_ONBOARD);
+        tx = is_onboarded();
+        break;
+    case RSK_WIPE: //--- wipe and onboard device ---
+        reset_if_starting(RSK_META_CMD_UIOP);
+        tx = onboard_device(&onboard_ctx);
+        clear_pin();
+        if (mode == BOOTLOADER_MODE_ONBOARD) {
+            *onboard_performed = 1;
+        }
```
amendelzon, line 106: Do we need the conditional here?

ledger/src/ui/src/bootloader.c
```diff
@@ -0,0 +1,227 @@
...
+            TRY {
+                rx = tx;
+                tx = 0; // ensure no race in catch_other if io_exchange throws
+                        // an error
+                rx = io_exchange(CHANNEL_APDU, rx);
+
+                tx = bootloader_process_apdu(rx, mode, &onboard_performed);
+                THROW(APDU_OK);
+            }
+            CATCH(EX_BOOTLOADER_RSK_END) {
```
amendelzon, line 216: Nice :)

ledger/src/ui/test/bootloader/Makefile
```diff
@@ -0,0 +1,48 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+SRCDIR = ../../src
+MOCKDIR = ./mock
+CFLAGS = -I$(MOCKDIR) -I$(SRCDIR) -g
+CFLAGS += -DPARAM_SIGNERS_FILE="testing"
```
amendelzon, line 26: Why do we need this?

ledger/src/ui/test/bootloader/mock/mock.h
```diff
@@ -0,0 +1,91 @@
...
+#ifndef _MOCK_H
+#define _MOCK_H
+
+#include <assert.h>
+#include <stdint.h>
+#include "os_exceptions.h"
+
+#define CHANNEL_APDU 0
+#define INS_ATTESTATION 0x50
+#define INS_SIGNER_AUTHORIZATION 0x51
```
amendelzon, line 34: We should probably grab these from the actual values if we can.

ledger/src/ui/test/bootloader/mock/mock.h
```diff
@@ -0,0 +1,91 @@
...
+#define IO_APDU_BUFFER_SIZE (5 + 255)
+extern unsigned char G_io_apdu_buffer[IO_APDU_BUFFER_SIZE];
+
+// Assert helpers
+#define ASSERT_EQUALS(a, b) assert((a) == (b))
+#define ASSERT_TRUE(cond) assert(cond)
+#define ASSERT_FALSE(cond) assert(!(cond))
+#define ASSERT_NOT_NULL(obj) assert(NULL != obj)
+#define ASSERT_APDU_AT(i, val) assert(APDU_AT(i) == val)
+#define ASSERT_FAIL() assert(false)
```
amendelzon, line 45: This is good, but I think we could benefit from richer helpers. So for example, instead of

```
ASSERT_APDU_AT(0, 0xAA);
ASSERT_APDU_AT(1, 0xBB);
ASSERT_APDU_AT(2, 0xCC);
```

we could have:

```
assert_apdu("aabbcc");
```

Also we don't necessarily need macros here, we are free to do whatever we want. I also don't think `ASSERT_EQUALS` and all of the very simple helpers really provide much of an advantage (e.g., `assert(!!cond)` or `assert(!cond)` are readable enough by themselves).

Lastly I think we could also add helpers for tasks like setting the APDU from a string, and we could also place all these helpers in a common directory so that we can share them across tests.

amendelzon (approved):
Looking great! Very minor comments on the code and some suggestions regarding the testing framework.

ledger/src/ui/test/bootloader/mock/mock.h
italo-sampaio: These values are defined in` attestation.h` / `signer_authorization.h`, adding these files would introduce a lot of dependencies (like the build flag mentioned on the comment above, private and public key type definitions, and we would need the implementation for all public functions in these modules). It is possible to add this, but I feel like this would make this test even more bloated than it is now.

If we ever change these defines, the tests will fail and we just need to update the definitions accordingly. What do you think?

ledger/src/ui/test/bootloader/Makefile
italo-sampaio: Not needed in the current implementation, but would be needed if we chose to include `signer_authorization.h` (see comment below for an explanation)

ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -0,0 +1,575 @@
...
+    G_pin_buffer_updated = true;
+    return 3;
+}
+
+unsigned int is_onboarded() {
+    SET_APDU_AT(1, G_is_onboarded);
+    SET_APDU_AT(2, VERSION_MAJOR);
+    SET_APDU_AT(3, VERSION_MINOR);
+    SET_APDU_AT(4, VERSION_PATCH);
+    return 5;
```
amendelzon, line 96: This looks like the actual implementation. Do we need to do this?

ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -0,0 +1,575 @@
...
+    G_unlock_called = false;
+    G_autoexec = 0;
+    G_reset_attestation_called = false;
+    G_reset_signer_authorization_called = false;
+    G_reset_onboard_called = false;
+}
+
+// Mock function calls
+unsigned int set_host_seed(volatile unsigned int rx, onboard_t* onboard_ctx) {
+    ASSERT_NOT_NULL(onboard_ctx);
```
amendelzon, line 81: Could we effectively check that this is called with the correct pointer?

ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -0,0 +1,575 @@
...
+    return rx;
+}
+
+unsigned int get_mode() {
+    SET_APDU_AT(1, RSK_MODE_BOOTLOADER);
+    return 2;
+}
+
+unsigned int get_attestation(volatile unsigned int rx, att_t* att_ctx) {
+    ASSERT_NOT_NULL(att_ctx);
```
amendelzon, line 123: Same question re the context pointer as before.

ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -0,0 +1,575 @@
...
+    G_get_attestation_called = true;
+    return 3;
+}
+
+unsigned int do_authorize_signer(volatile unsigned int rx,
+                                 sigaut_t* sigaut_ctx) {
+    ASSERT_NOT_NULL(sigaut_ctx);
```
amendelzon, line 130: And same here re the sigaut context pointer.

ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -0,0 +1,575 @@
...
+void reset_onboard_ctx(onboard_t* onboard_ctx) {
+    G_reset_onboard_called = true;
+}
+
+// Function definitions required for compiling bootloader.c
+void init_signer_authorization() {
+}
+
+unsigned short io_exchange(unsigned char channel, unsigned short tx_len) {
+    return 0;
```
amendelzon, line 166: Since we're here, we could check this is called with the correct parameters (namely, the correct `channel`).

ledger/src/ui/test/mock/apdu_utils.h
```diff
@@ -0,0 +1,39 @@
...
+#include <string.h>
+
+#define CHANNEL_APDU 0
+
+#define IO_APDU_BUFFER_SIZE (5 + 255)
+extern unsigned char G_io_apdu_buffer[IO_APDU_BUFFER_SIZE];
+
+#define ASSERT_APDU(str) \
+    assert(0 == strncmp((const char*)G_io_apdu_buffer, str, strlen(str)))
+#define ASSERT_APDU_AT(i, val) assert(APDU_AT(i) == val)
```
amendelzon, line 36: These are great. Just a couple of questions:
- Is there really a need for `ASSERT_APDU_AT`? Couldn't we just assert the entirety of the APDU every time?
- Why are we using `#define` as opposed to proper functions? I'm thinking functions tend to be easier to read and modify for e.g. debugging purposes.

amendelzon (commented):
Things are looking good! A few questions and comments.

ledger/src/ui/test/mock/apdu_utils.h
italo-sampaio: Good point about `ASSERT_APDU_AT`. My reasoning is that for the purpose of these tests, we don't always need to check the entire APDU buffer, so sometimes it should be useful to check only one specific position. But giving this a second thought, asserting the full buffer every time probably adds completion to the tests. I'll change that.

About using `defines`, I reckon that using functions has several advantages. The only thing that bothers me about using functions is that in the event of an error, the error message will point to the line where `assert()` was actually called, inside the helper function. So any failed assert will point to the same line of `assert_utils.c`. That is less helpful for debugging IMO. When using defines, the error message points to the line where the macro was called, which lets you know exactly which assert failed. What are your thoughts on that?

amendelzon: Good point about the lines in the error messages.

amendelzon (approved):
Looks good to me, great work, thank you!

PR #109, opened 2023-01-03
Added UI attestation module unit tests

italo-sampaio:


ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,359 @@
...
+    reset_attestation(&G_att_ctx);
+    ASSERT_STRUCT_CLEARED(att_t, G_att_ctx);
+    assert(att_stage_wait_ud_value == G_att_ctx.stage);
+}
+
+void test_get_attestation_ud_value() {
+    printf("Test ATT_OP_UD_VALUE...\n");
+    reset_attestation(&G_att_ctx);
+    *(unsigned char *)N_onboarded_ui = 1;
+    memcpy(&G_att_ctx.priv_key_data, PRIVATE_KEY, sizeof(PRIVATE_KEY));
```
amendelzon, line 152: Same here with `priv_key_data` and `&`.

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,359 @@
...
+    G_att_ctx.stage = att_stage_wait_ud_value;
+    // CLA + INS_ATTESTATION + ATT_OP_UD_VALUE + UD_VALUE
+    set_apdu("\x80\x50\x01\x46\x8d\xa8\x7f\x6a\x85\xe6\x40\x93\x27\xe1\x17\xe8"
+             "\xc7\xd2\x11\x0c\x73\x60\x22\x26\xbb\xb5\xed\xf2\x7d\x98\xc8\xa3"
+             "\x1b\xcc\xf0");
+
+    assert(3 == get_attestation(35, &G_att_ctx));
```
amendelzon, line 159: Can we get rid of the magic number here and use the return value of `set_apdu` instead?

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,359 @@
...
+    }
+    END_TRY;
+}
+
+void test_get_attestation_get_msg() {
+    printf("Test ATT_OP_GET_MSG...\n");
+    reset_attestation(&G_att_ctx);
+    *(unsigned char *)N_onboarded_ui = 1;
+    memcpy(
+        &G_att_ctx.msg,
```
amendelzon, line 204: Same here with `msg` and `&`.

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,359 @@
...
+    ASSERT_APDU(
+        "\x80\x50\x02\x00"
+        "HSM:UI:3.0"
+        "\x46\x8d\xa8\x7f\x6a\x85\xe6\x40\x93\x27\xe1\x17\xe8\xc7\xd2\x11\x0c"
+        "\x73\x60\x22\x26\xbb\xb5\xed\xf2\x7d\x98\xc8\xa3\x1b\xcc\xf0"
+        "\x03\xe6\xd7\x1d\x5c\x2b\x06\x36\x03\x53\xfb\xd8\x22\x7a\xb3\xab\xfc"
+        "\x3d\x46\x6a\x5f\x74\xdc\x28\xc2\xb7\x3e\xb0\x95\x2b\xec\x20\x87"
+        "\xdd\xa7\x70\x05\x55\xa3\x7b\x75\x34\x29\x1b\x96\x2d\x9f\x41\x41\xb9"
+        "\x64\x48\xda\xd7\x12\x81\xef\x7c\x2d\x61\x49\x4c\xcb\xb8\x59"
+        "\x00\x09");
```
amendelzon, line 231: This is 113 bytes, and therefore longer than our actual APDU, which is ~85 bytes. The idea of getting the message with paging capabilities is that we can cope with the small APDU. Can we write this to be able to test the paging code maybe using a smaller APDU?

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,359 @@
...
+void test_get_attestation_get() {
+    printf("Test ATT_OP_GET...\n");
+    reset_attestation(&G_att_ctx);
+    *(unsigned char *)N_onboarded_ui = 1;
+    G_att_ctx.stage = att_stage_ready;
+
+    // CLA + INS_ATTESTATION + ATT_OP_GET
+    set_apdu("\x80\x50\x03");
+
+    assert(35 == get_attestation(3, &G_att_ctx));
```
amendelzon, line 278: Should we be checking for the correct signature here? Is this checked somewhere else? Also where does the `35` come from? Can we use a constant or something more declarative?

amendelzon (changes requested):
Looking good so far. I think there might be some issues with pointer usage, and also a need for additional testing of some scenarios.

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,359 @@
...
+}
+
+// cx mocks
+int cx_ecdsa_init_private_key(cx_curve_t curve,
+                              unsigned char *rawkey,
+                              unsigned int key_len,
+                              cx_ecfp_private_key_t *key) {
+    assert(CX_CURVE_256K1 == curve);
+    ASSERT_STR_N_EQUALS(rawkey, PRIVATE_KEY, KEYLEN);
+    assert(rawkey == (unsigned char *)&G_att_ctx.priv_key_data);
```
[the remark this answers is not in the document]
italo-sampaio: Nice catch! Oddly enough, I double-checked and both `G_att_ctx.priv_key_data` and `&G_att_ctx.priv_key_data` return the same address, that's why the assert still works. That makes sense, but the compiler should at least issue a warning when you apply the `&` operator to a static array IMO.

Anyway, I'm fixing this, thanks for pointing it out!

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -216,19 +217,23 @@ void test_get_attestation_get_msg() {
...
         "\x3d\x46\x6a\x5f\x74\xdc\x28\xc2\xb7\x3e\xb0\x95\x2b\xec\x20\x87"
-        "\xdd\xa7\x70\x05\x55\xa3\x7b\x75\x34\x29\x1b\x96\x2d\x9f\x41\x41\xb9"
-        "\x64\x48\xda\xd7\x12\x81\xef\x7c\x2d\x61\x49\x4c\xcb\xb8\x59"
-        "\x00\x09");
+        "\xdd\xa7\x70\x05");
+
+    // CLA + INS_ATTESTATION + ATT_OP_GET_MSG + PAGE_NUM (1)
+    set_apdu("\x80\x50\x02\x01");
+    assert(34 == get_attestation(4, &G_att_ctx));
+    ASSERT_APDU("\x80\x50\x02\x00"
```
amendelzon, line 233: This seems odd. If there are more pages, then the response at the fourth byte should have an `\x01` indicating that. Can you double check please? (see `attestation.c:274`). Maybe an actual bug?

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -216,19 +217,23 @@ void test_get_attestation_get_msg() {
 
     // CLA + INS_ATTESTATION + ATT_OP_GET_MSG + PAGE_NUM (0)
     set_apdu("\x80\x50\x02\x00");
-
-    int ret = get_attestation(4, &G_att_ctx);
-    assert((ATT_MESSAGE_SIZE + 4) == get_attestation(4, &G_att_ctx));
+    assert((APDU_TOTAL_DATA_SIZE_OUT + 3) == get_attestation(4, &G_att_ctx));
     ASSERT_APDU(
```
amendelzon, line 221: This seems odd. If there are more pages, then the response at the fourth byte should have an `\x01` indicating that. Can you double check please? (see `attestation.c:274`). Maybe an actual bug?

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,385 @@
...
+    assert(att_stage_wait_ud_value == G_att_ctx.stage);
+}
+
+void test_get_attestation_ud_value() {
+    printf("Test ATT_OP_UD_VALUE...\n");
+    unsigned int rx;
+
+    reset_attestation(&G_att_ctx);
+    *(unsigned char *)N_onboarded_ui = 1;
+    memcpy(G_att_ctx.priv_key_data, PRIVATE_KEY, sizeof(PRIVATE_KEY));
```
amendelzon, line 154: This is part of the behaviour we are testing, so we shouldn't be doing it as part of the test setup. Sorry, I didn't spot it in the previous review. This should be done by the derivation function from the path.

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,385 @@
...
+                                                  unsigned char *signature) {
+    assert(src == (unsigned char *)G_att_ctx.msg);
+    assert(srcLength == G_att_ctx.msg_offset);
+    assert(signature == APDU_DATA_PTR);
+    memcpy(signature, MSG_SIGNATURE, sizeof(MSG_SIGNATURE) - 1);
+
+    return sizeof(MSG_SIGNATURE) - 1;
+}
+
+void os_perso_derive_node_bip32(cx_curve_t curve,
```
amendelzon, line 110: This is where the private key data should be set (see my next comment).

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -0,0 +1,385 @@
...
+    "\x46\x6a\x5f\x74\xdc\x28\xc2\xb7\x3e\xb0\x95\x2b\xec\x20\x87\x89\x35\xa7" \
+    "\xe5\x6c\x84\xd0\xb2\xad\xae\x28\x58\x40\xd0\x24\xd5\x12\x6c\x7a\x3b\x6c" \
+    "\x34\xc5\x22\x89\x42\x0b\x32\x08\xf2\x4d"
+#define PRIVATE_KEY                                                            \
+    "\xce\x1c\xa0\xf8\xf6\x40\x4b\xfd\x7a\x0a\x33\x80\xd3\x8c\x21\x5c\x89\x31" \
+    "\x61\x91\xb5\x45\x62\xd8\x32\x68\xca\x57\x01\x73\x30\xfc"
+#define SIGNER_HASH                                                            \
+    "\xdd\xa7\x70\x05\x55\xa3\x7b\x75\x34\x29\x1b\x96\x2d\x9f\x41\x41\xb9\x64" \
+    "\x48\xda\xd7\x12\x81\xef\x7c\x2d\x61\x49\x4c\xcb\xb8\x59"
+#define MSG_SIGNATURE                                                          \
```
amendelzon, line 42: Signatures are usually around 70 bytes long. Maybe do 75 just to make sure bigger messages fit. Sorry, didn't catch it on my previous review.

amendelzon (changes requested):
Looking good! Minor things that we need to address and then we're ready to merge.

ledger/src/ui/test/attestation/test_attestation.c
italo-sampaio: Updated to 75 bytes, thanks for noticing!

ledger/src/ui/test/attestation/test_attestation.c
italo-sampaio: fixed!

amendelzon (approved):
LGTM!

PR #110, opened 2023-01-04
Session-persist partial advance blockchain state

amendelzon:
Moving partial advance blockchain state to an independent memory area
    
- Preventing other operations from resetting partial advance blockchain state (as when in NVM)
- Any error conditions still reset the partial advance blockchain state
- Adding new tests
- Renaming test suite files to accomodate for increasing number of tests


italo-sampaio (approved):
LGTM!

PR #112, opened 2023-01-05
Improvements on buffer comparison on UI unit tests

italo-sampaio:
- Make sure we don't treat binary data as strings for data comparison
- Use `memcpy` for implementation of `set_apdu`


ledger/src/ui/test/bootloader/test_bootloader.c
```diff
@@ -190,7 +190,8 @@ void test_seed() {
     reset_flags();
     G_bootloader_mode = BOOTLOADER_MODE_DEFAULT;
     G_host_seed_is_set = false;
-    rx = set_apdu("\x80\x44"); // RSK_SEED_CMD
+    set_apdu("\x80\x44", 2); // RSK_SEED_CMD
```
amendelzon, line 193: Maybe you can define a macro to call this function with a string literal using `sizeof()-1`? Same for the assertion one. It should only be used with string literals tho, so would try to make that clear if you do. You could also add `rx` as a parameter and set it within the macro.

ledger/src/ui/test/mock/assert_utils.h
```diff
@@ -33,4 +33,5 @@
     assert(0 == strcmp((const char *)a, (const char *)b))
 #define ASSERT_STR_N_EQUALS(a, b, n) \
```
amendelzon, line 34: Are we still using this one? We probably shouldn't if we are unless there's good reason to.

ledger/src/ui/test/onboard/test_onboard.c
```diff
@@ -76,8 +76,11 @@ unsigned int bolos_ux_mnemonic_from_data(unsigned char *in,
                                          unsigned int inLength,
                                          unsigned char *out,
                                          unsigned int outLength) {
-    sprintf((char *)out, "mnemonics-generated-from-%s", in);
-    return strlen((const char *)out);
+    int len = strlen("mnemonics-generated-from-");
```
amendelzon, line 79: We should also bear in mind that string literals in C are automatically appended a null-terminator.

amendelzon (commented):
Looking good. Left some comments, maybe still some room for improvement?

amendelzon (approved):
LGTM!

PR #115, opened 2023-01-13
Prevent re-onboarding

amendelzon:
- Preventing SEED and WIPE commands from executing if device is already onboarded
- Preventing onboarding tool from executing onboarding if device is already onboarded
- Added unit test cases for firmware and middleware changes
- Improving output for `HSM2DongleErrorResult` errors


italo-sampaio (approved):
LGTM!

PR #116, opened 2023-01-13
Added unit tests for UI signer authorization module

italo-sampaio:
- Added signer authorization unit tests
- Added signer authorization unit tests to CI
- Added `NON_VOLATILE` modifier to NVM variables
- Moved `runtime.h` to common folder to allow usage in UI
- Incidentally fixing log message on Makefile


ledger/src/ui/src/signer_authorization.c
```diff
@@ -50,7 +51,7 @@ static const uint16_t INITIAL_SIGNER_ITERATION = PARAM_INITIAL_SIGNER_ITERATION;
 #endif
 
 // Current signer status
-static const sigaut_signer_status_t N_current_signer_status_var;
+static NON_VOLATILE sigaut_signer_status_t N_current_signer_status_var;
```
amendelzon, line 54: We could potentially remove the `static` in order to be able to use `extern` from unit tests. I don't see a lot of harm in it, do you?

ledger/src/ui/test/signer_authorization/test_signer_authorization.c
```diff
@@ -0,0 +1,449 @@
...
+}
+
+int cx_ecfp_init_public_key(cx_curve_t curve,
+                            unsigned char *rawkey,
+                            unsigned int key_len,
+                            cx_ecfp_public_key_t *key) {
+    assert(NULL != rawkey);
+    assert(NULL != key);
+    assert(CX_CURVE_256K1 == curve);
+    assert(65 == key_len);
```
amendelzon, line 124: Isn't the raw key here meant to be the private key? In that case, it would be 32 bytes long.

ledger/src/ui/test/signer_authorization/test_signer_authorization.c
```diff
@@ -0,0 +1,449 @@
...
+
+void test_op_sigaut_sigver() {
+    printf("Test OP_SIGAUT_SIGVER...\n");
+
+    unsigned int rx;
+    reset_signer_authorization(&G_sigaut_ctx);
+    G_sigaut_ctx.state = sigaut_state_wait_signer_version;
+
+    // OP_SIGAUT_SIGVER + hash + iteration
+    SET_APDU("\x80\x81\x01" PARAM_INITIAL_SIGNER_HASH "\x00\x02", rx);
```
amendelzon, line 200: You probably want to try and use a different signer hash to test this since it'll usually be the case, right?

ledger/src/ui/test/signer_authorization/test_signer_authorization.c
```diff
@@ -0,0 +1,449 @@
...
+    // OP_SIGAUT_SIGN + G_valid_signature[1]
+    SET_APDU("\x80\x81\x02\xf9\x6f\x6b\x3e\x3c\xe5\x08\x8b\x46\x83\xbe\x87\x3c"
+             "\x07\x5a\x61\x4a\x8c\x65\x64\x9a\x4c\xa8\x6a\x78\xa0\xc6\x6c\x9e"
+             "\x47\x7e\x20\x5f\x32\x2e\xc9\x89\x9e\x0f\x4d\x00\x60\xab\x30\x69"
+             "\x28\xcd\x69\x43\x01\x30\x36\x01\x33\xf1\x32\x26\xcd\x1c\x62\x3b"
+             "\x12\xf7\xf4\xef\x69\x87\x8b\xe8\x7d\xbe\x60\x86\x4d\xa2",
+             rx);
+    assert(4 == do_authorize_signer(rx, &G_sigaut_ctx));
+    assert(RES_SIGAUT_SUCCESS == APDU_DATA_PTR[0]);
+    assert_sigaut_ctx_reset(&G_sigaut_ctx);
```
amendelzon, line 306: Should we also test that the authorized signer version has effectively changed?

ledger/src/ui/test/signer_authorization/test_signer_authorization.c
```diff
@@ -0,0 +1,449 @@
...
+    assert(TX_FOR_DATA_SIZE(sizeof(G_authorizers_pubkeys[2])) ==
+           do_authorize_signer(rx, &G_sigaut_ctx));
+    ASSERT_MEMCMP(APDU_DATA_PTR,
+                  G_authorizers_pubkeys[2],
+                  sizeof(G_authorizers_pubkeys[2]));
+}
+
+// Note: this test trusts in values written to NVM during signer authorization,
+//       so it must be run after test_op_sigaut_sign
+void test_is_authorized_signer() {
```
amendelzon, line 413: Like I mentioned before, maybe we can remove the `static` keyword from the NVM variable to be able to manipulate it here, given that there's no unnecessary risks taken.

amendelzon (commented):
Looking good, left a few comments of things we might want to look into.

ledger/src/ui/test/signer_authorization/test_signer_authorization.c
italo-sampaio: Actually, this is the public key. All this function does is initialize a `cx_ecfp_public_key_t` struct with the given raw key. This function is called on` signer_authorization.c` as:

`cx_ecfp_init_public_key(CX_CURVE_256K1, (unsigned char*)authorizers_pubkeys[i], sizeof(authorizers_pubkeys[i]), &sigaut_ctx->pubkey);`

amendelzon: Cool, thanks for clarifying.

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -53,7 +54,7 @@ static cx_ecfp_private_key_t G_priv_key;
 static unsigned char G_path[PUBKEY_PATH_LENGTH];
 
 // Global onboarding flag
-const unsigned char N_onboarded_ui[1];
+NON_VOLATILE unsigned char N_onboarded_ui[1];
```
amendelzon, line 57: Does using this macro here make sense at all?

amendelzon (dismissed):
LGTM!
One minor comment regarding the use of `NON_VOLATILE` in test-only NVM mock variables.

ledger/src/ui/test/attestation/test_attestation.c
italo-sampaio: This is here just to keep the same definition as in `attestation.c`. Cold be removed and the generated code would be the same, but I think it looks better if it follows the original definition. What do you think?

amendelzon: Yes, I understand the motivation, but the bottomline is those are different pieces of code and a change in one side doesn't guarantee a change in the other. In other words it's an artificial replication, a copy, and it will need to be maintained. And I can see us totally missing out on that in the future.

Is there a way we can include the actual definition as opposed to just copying it?

italo-sampaio: Fair enough. Updated definition!

ledger/src/ui/test/attestation/test_attestation.c
```diff
@@ -27,6 +27,7 @@
 #include "apdu_utils.h"
 #include "assert_utils.h"
 #include "err.h"
+#include "runtime.h"
```
amendelzon, line 30: Do we need this?

amendelzon (dismissed):
LGTM. One minor comment

ledger/src/ui/test/attestation/test_attestation.c
italo-sampaio: Nice catch! Removed

amendelzon (approved):
LGTM!

PR #120, opened 2023-01-23
Added a binary build for signapp tool as part of the distribution

italo-sampaio:
Added a binary build for `signapp` tool as part of the distribution


amendelzon (commented):
Looking good. We should probably also update the `middleware/build/all` script.

amendelzon (approved):
LGTM!

PR #122, opened 2023-01-25
Changed signapp usage

italo-sampaio:
In the past, we changed `siggnapp` usage to use `-s` for the signer firmware instead of `-a`. We are now changing this back to `-a` to keep it backward compatible with older versions of our tools.


amendelzon (approved):
LGTM!

PR #123, opened 2023-01-26
Document deployment flags on UI and signer's Makefiles

italo-sampaio:
Added documentation for the values used for the `appFlags` argument on both UI and signer Makefiles

### DIFF
--- a/ledger/src/signer/Makefile
+++ b/ledger/src/signer/Makefile
@@ -47,7 +47,6 @@ APPNAME = "RSK Sign"
 APPVERSION = 3
 # No flags
 APPFLAGS = 0x00
-APP_LOAD_PARAMS = --appFlags $(APPFLAGS) $(COMMON_LOAD_PARAMS)
 PROG = "app"
 
 # Build configuration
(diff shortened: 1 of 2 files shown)


ledger/src/signer/Makefile
```diff
@@ -45,7 +45,9 @@ include $(BOLOS_SDK)/Makefile.defines
 # Main app configuration
 APPNAME = "RSK Sign"
 APPVERSION = 3
-APP_LOAD_PARAMS = --appFlags 0x00 $(COMMON_LOAD_PARAMS)
+# No flags
+APPFLAGS = 0x00
+APP_LOAD_PARAMS = --appFlags $(APPFLAGS) $(COMMON_LOAD_PARAMS)
```
amendelzon, line 50: Are we actually using `APP_LOAD_PARAMS`? Maybe we could remove it if we're not.

amendelzon (approved):
LGTM! One minor suggestion.

ledger/src/signer/Makefile
italo-sampaio: Nice catch! removed!

amendelzon (approved):
LGTM!

PR #124, opened 2023-02-01
UI heartbeat

amendelzon:
- Implemented UI heartbeat with its own main loop
- Implemented UI heartbeat operation in middleware
- UI dashboard has now got three action modes instead of two (added UI heartbeat main loop)
- Factored out UI I/O exchange exception handling into the communication module
- Factored out signer authorization NVM status to a separate module (to allow for TCPSigner implementation)
- Renamed UI constants to have a different prefix (to allow for TCPSigner implementation)
- Renamed bolos_ux_handlers to ux_handlers to avoid confusion with BOLOS code
- Added UI heartbeat capabilities to the TCPSigner
- Renamed UI sources to prevent clashing when implementing TCPSigner UI heartbeat
- Added exit functionality to signer
- NVM backing up of the partial advance blockchain state for signer close/open scenarios
- Resetting signer heartbeat state in some error scenarios
- Smaller signer heartbeat max message size additional compile time checks
- Factored out operating modes into common code
- Factoring out general constants into common code
- Any signer is authorized when UI is built in debug mode
- Added an option to exit the signer to the reconnect Signer test case runner, updated test cases
- Added UI unit tests and fixed existing
- Added middleware UI heartbeat unit tests and fixed existing
- Updated protocol documentation
- Added UI heartbeat documentation
- Adding scripts to aid running ledger unit tests, updating github workflow


ledger/src/tcpsigner/tcpsigner.c
```diff
@@ -60,6 +61,12 @@ typedef enum {
     ARG_NU_IRIS,
 } arg_non_printable_t;
 
+/**
+ * UI heartbeat memory area
+ * There's probably a better
```
italo-sampaio, line 66: This comment looks incomplete, should it be "there's probably a better **place**"?

ledger/src/ui/src/communication.c
```diff
@@ -22,44 +22,89 @@
...
+ * This returns either bootloader or heartbeat mode, depending
+ * on the argument (which signals where it is called from)
  *
- * @ret number of transmited bytes to the host
+ * @arg[in] ui_heartbeat_main whether called from the ui heartbeat main
+ * @returns number of transmited bytes to the host
  */
-unsigned int get_mode() {
+unsigned int get_mode(bool ui_heartbeat_main) {
     unsigned char output_index = CMDPOS;
```
italo-sampaio, line 50: This function always returns either `APP_MODE_UI_HEARTBEAT` or `APP_MODE_BOOTLOADER`, depending on where this was called from. This logic of a function having different behaviors depending on where it was called from is confusing. Also, it is hard to understand what `get_mode(false)` actually means just by looking at the code. What do you think of breaking this into two different functions, something along the lines of:

```
unsigned int report_mode_bootloader()
unsigned int report_mode_heartbeat()
```

italo-sampaio (changes requested):
Looking good! Just one incomplete comment and one suggested change, and this should be good to go!

ledger/src/ui/src/communication.c
amendelzon: Good point, fixed!

ledger/src/tcpsigner/tcpsigner.c
amendelzon: Good catch, fixed!

italo-sampaio (approved):
LGTM!

PR #126, opened 2023-02-02
Added static code analysis for ledger code

italo-sampaio:
- Added clang static code analysis to signer
- Added clang static code analysis to UI


amendelzon (commented):
Looks good! I would also add the output of the analysis to the corresponding `.gitignore`.

PR #128, opened 2023-02-08
Fixed varint handling on BTC tx parser and RSK trie parser

amendelzon:
- Handling partial varint parsing in BTC tx parsing
- Fixing an issue when accumulating a varint raw data both in BTC tx parsing and RSK trie parsing
- Added both BTC tx and trie unit test cases
- Added an authorized signing test case to the signer tests
- Incidentally fixing a log message format


italo-sampaio (approved):
LGTM!

PR #131, opened 2023-03-16
Minor signer and UI tweaks

amendelzon:
- Zeroing out sRLP context when initialising the parser
- Changing deployed UI and Signer version to 4
- Removed unused signer memory `#define`s


italo-sampaio (approved):
Looks good to me!

PR #137, opened 2023-05-11
Updated base image for mware, tcpsigner-bundle and packer

italo-sampaio:
The previously used base image, debian stretch, has reached end of life. This was causing build issues both locally and on the CI. This commit updates the base image to debian bullseye, which is the current stable release with LTS. This also updates some package versions to the ones currently supported.


amendelzon (approved):
LGTM!

PR #139, opened 2023-05-17
Updated base image for powHSM distribution

italo-sampaio:
Following the middleware, tcpsigner-bundle and packer docker image updates we also need to update the base image for the powHSM distribution in order to be able to build and run the generated distributions.


amendelzon (approved):
LGTM!

PR #147, opened 2023-05-30
Using nanos SDK custom fork

amendelzon:
- Removing now unnecessary patches to the SDK in the Dockerfile
- Reduced number of max USB endpoints for UI and Signer
- Updated CHANGELOG


italo-sampaio (approved):
LGTM!

PR #152, opened 2023-08-01
Fixing middleware docker image build

amendelzon:
- Fix middleware docker image build: installing an older Cython before the rest of the dependencies. This prevents some package builds from failing.


italo-sampaio (approved):
Looks good to me!

Just one side note: does it make sense to bump the version since we are not touching the firmware, just the build environment?

amendelzon:
> Looks good to me!
> 
> Just one side note: does it make sense to bump the version since we are not touching the firmware, just the build environment?

That's a good question. We do want to avoid overwriting tags. What would be a good alternative?

italo-sampaio:
> > Looks good to me!
> > Just one side note: does it make sense to bump the version since we are not touching the firmware, just the build environment?
> 
> That's a good question. We do want to avoid overwriting tags. What would be a good alternative?

Fair point on not overwriting tags. I also prefer the risk of adding too much information to the risk of having the same tag with two different meanings, so I would say let's merge it with the version bumping since it does no harm.

italo-sampaio (approved):
LTGM!

PR #153, opened 2023-08-02
Fixing middleware docker image build

amendelzon:
Fix middleware docker image build: installing an older Cython before the rest of the dependencies. This prevents some package builds from failing.


italo-sampaio (approved):
LGTM

PR #158, opened 2023-10-23
Fixed failing middleware docker image build

amendelzon:


italo-sampaio (approved):
Looks good and passed the middleware docker build stage on Checks, good to go!

PR #165, opened 2023-12-14
Added firmware static analysis to CI

italo-sampaio:
- Merged old static analysis scripts into a single script
- Added static analysis to the CI on pushes to master
- Generated report is available for download on job summary


italo-sampaio:
An example run can be seen [here](https://github.com/rsksmart/rsk-powhsm/actions/runs/7465070272).

amendelzon (dismissed):
LGTM!

As a question: I wonder whether it would make a difference using the arm clang toolchain we use for building firmware for the Nano S instead of the system /usr/bin/clang*.

amendelzon (approved):
LGTM!

PR #177, opened 2024-05-21
Decoupling business layer from hardware layer in Signer component

amendelzon:
- Added HAL headers
- Added HAL implementation for Ledger
- Refactored necessary common sources
- Refactored necessary Signer sources
- Updated Signer Makefile to include HAL code
- Removed HSM_SIMULATOR define in favor of HSM_PLATFORM_* defines
- Included X86 (partial) implemenations of some HAL modules (exceptions, logging, hashing)
- Renamed UI communication module to avoid name clashing with HAL
- Changed platform define in UI Makefile
- Added linting to HAL

All physical Ledger tests currently passing.


italo-sampaio (approved):
LGTM!

PR #179, opened 2024-05-29
Decoupling business layer from hardware layer in TCPSigner

amendelzon:
- Implementing HAL for x86 based on the existing TCPSigner implementation
- Additional x86-only functions for HAL implementation
- Updating TCPSigner implementation to use HAL
- Added TCPSigner-specific UI dependencies to implement functions needed by UI only code
- Factoring HAL constants into a constants file
- New BIP32 parser for TCPSigner seed file
- Ensuring Firmware tests pass both on Ledger Nano S and TCPSigner
- Linting


italo-sampaio (approved):
Looks good to me!

PR #184, opened 2024-06-27
Upgrading Python to version 3.12

amendelzon:
- Updated docker images to use a Debian Bookworm based Python 3.12 base image
- Updated `hidapi` repository reference to a version compatible with the latest Cython
- Updated Python libraries
- Unified dependencies in `requirements.txt` (except for `hidapi`, which is manually installed)
- Added one-round `keccak256` wrapper that uses `pycryptodome`, and corresponding unit tests
- Replaced uses of `pysha3` with new `keccak256` wrapper
- Removed `pysha3` from Python libraries
- Fixed existing unit tests
- Replaced TCPSigner use of deprecated `secp256k1_ec_privkey_tweak_add` with the newer `secp256k1_ec_seckey_tweak_add`
- Fixed new Python linting findings
- Updated integration tests tag


italo-sampaio (approved):
Awesome work, looks good!

PR #189, opened 2024-07-25
Moving I/O (aka communication) layer outside powHSM code

amendelzon:
- The powHSM logic now handles only one command at a time
- Each implementation (i.e., Signer and TCPSigner) implement their own I/O and looping
- The communication module only facilitates access to the APDU buffer


firmware/src/hal/src/x86/communication.c
```diff
@@ -22,404 +22,23 @@
  * IN THE SOFTWARE.
  */
 
-#include <stdio.h>
```
italo-sampaio, line 25: How about moving this to `firmware/src/hal/src/common/communication.c` since it is now completely platform-independent?

firmware/src/ledger/signer/src/main.c
```diff
@@ -210,6 +210,34 @@ unsigned char io_event(unsigned char channel) {
...
+    }
+    END_TRY;
+}
+
+static void main_loop() {
+    volatile unsigned int rtx = 0;
+
+    while (!hsm_exit_requested()) {
+        if (!do_io_exchange(&rtx))
+            continue;
```
italo-sampaio, line 236: Aren't we ignoring a potential exception that could happen in `do_io_exchange` here?

italo-sampaio (commented):
Looks good overall. Left a few questions inline, let me know your thoughts.

firmware/src/ledger/signer/src/main.c
amendelzon: I don't quite follow. Ignoring it how?

italo-sampaio: Oh, you're right. I thought the exception would be handled by `hsm_process_apdu`, which we're skipping, but it will be sent to the host in the next `do_io_exchange` call.

All good, never mind. I'll approve the PR, since my other comment is just an optional suggestion.

italo-sampaio (approved):

firmware/src/ledger/signer/src/main.c
amendelzon: Oh, gotcha. Thanks for clarifying!

firmware/src/hal/src/x86/communication.c
amendelzon: We are actually currently just using symlinks to the `firmware/src/hal/src/common` directory, so will do that. I won't be removing symlinks altogether since what is going to happen for SGX is that we will share some libraries between platforms, but not between all platforms (e.g., we might share `keccak256` between SGX and x86, but Ledger will have its own). So most likely the `common` directory will just have to go.

italo-sampaio (approved):
Looks good!

PR #191, opened 2024-08-08
Improving HAL directory structure

amendelzon:
To accomodate for testing different platform implementations of same-name units. Also updated corresponding makefiles.


italo-sampaio (approved):
Looking good!

PR #196, opened 2024-09-04
SGX simulation build

amendelzon:
- Added simulator-only seal and unseal implementations (seal/unseal is unsupported in simulation mode and errors out)
- Added simulation build script
- Added separate simulation configuration file


italo-sampaio (approved):
Nice addition, looks good!

PR #197, opened 2024-09-04
Fix verify_attestation.py to accept distinct versions for UI and Signer

italo-sampaio:
- Version MINOR is now allowed to be distinct between UI and Signer
- Script outputs the installed version of both apps

Example output after a simulated signer upgrade to `5.2`:

```
########################################
### -> Verify UI and Signer attestations
########################################
Using 0490f5c9d15a0134bb019d2afd0bf297149738459706e7ac5be4abc350a1f818057224fce12ec9a65de18ec34d6e8c24db927835ea1692b14c32e9836a75dad609 as root authority
--------------------------------------------------------------------------------------------------------
UI verified with:
UD value: 474cc830bb9159299ca788a7527b561817ffa42f5d13f5429f14a95b0ee154f0
Derived public key (m/44'/0'/0'/0/0): 0332aba656b37e4e927806e1fe72323e1a7ca784fcc0eb9de2b1590f32f822b956
Authorized signer hash: 9f21e535d64ae9ab768fb2b6ab43e645c4d78a18318dca656de9c329f26098e9
Authorized signer iteration: 2
Installed UI hash: f060858f70c1832b2ac06c88fd741a4c8a098c974ad51122b2669d57c1f9d89c
Installed UI version: 5.1
--------------------------------------------------------------------------------------------------------
---------------------------------------------------------------------------------------
Signer verified with public keys:
m/44'/0'/0'/0/0:   0332aba656b37e4e927806e1fe72323e1a7ca784fcc0eb9de2b1590f32f822b956
m/44'/1'/0'/0/0:   03db12ee740f88bb412b66aaed99bd1908917d9321f6393643d1435c6b59f8ee4f
m/44'/1'/1'/0/0:   031867a30ec059b17d5901092b866d81fcf359fb528954c1ece903d1667a114a22
m/44'/1'/2'/0/0:   02078a43a98b19be5c4cacf2cc6ae2e2431de20561540ba40b344e8884ddc86816
m/44'/137'/0'/0/0: 028047542a7bc3007eed06e8e690d5447d89a5d394a351b189e33c40545afcbcd3
m/44'/137'/1'/0/0: 02bbf36ba664cfa81c659aceadc50c6b28462cb720b5dd3ec5afd8c8bd593ce5a8

Hash: db8fcfb3827812d17a87e645ac19469469e9f9626608cc33a5aa4d361d9dc3a8
Installed Signer hash: 9f21e535d64ae9ab768fb2b6ab43e645c4d78a18318dca656de9c329f26098e9
Installed Signer version: 5.2
---------------------------------------------------------------------------------------
```

### DIFF
--- a/middleware/admin/verify_attestation.py
+++ b/middleware/admin/verify_attestation.py
@@ -23,6 +23,7 @@
 import json
 import hashlib
 import secp256k1 as ec
+import re
 from .misc import info, head, AdminError
 from .utils import is_nonempty_hex_string
 from .certificate import HSMCertificate
@@ -46,23 +47,13 @@ DEFAULT_ROOT_AUTHORITY = "0490f5c9d15a0134bb019d2afd0bf297149738459706e7ac5be4ab
 
 
 def validate_ui_message_header(ui_message):
-    minor_offset = len(UI_MESSAGE_HEADER) - 1
-    if ui_message[:minor_offset] != UI_MESSAGE_HEADER[:minor_offset]:
-        raise AdminError()
-    version_minor = ui_message[minor_offset]
-    # The minor version must be a single digit between 0 and 9
-    if version_minor < 48 or version_minor > 57:
-        raise AdminError()
+    header = ui_message[:len(UI_MESSAGE_HEADER)]
+    return re.compile(b"^HSM:UI:5.[0-9]$").match(header) is not None
 
 
 def validate_signer_message_header(signer_message):
-    minor_offset = len(SIGNER_MESSAGE_HEADER) - 1
-    if signer_message[:minor_offset] != SIGNER_MESSAGE_HEADER[:minor_offset]:
-        raise AdminError()
-    version_minor = signer_message[minor_offset]
-    # The minor version must be a single digit between 0 and 9
-    if version_minor < 48 or version_minor > 57:
-        raise AdminError()
+    header = signer_message[:len(SIGNER_MESSAGE_HEADER)]
+    return re.compile(b"^HSM:SIGNER:5.[0-9]$").match(header) is not None
 
 
 def do_verify_attestation(options):
@@ -142,13 +133,12 @@ def do_verify_attestation(options):
     ui_message = bytes.fromhex(ui_result[1])
     ui_hash = bytes.fromhex(ui_result[2])
     mh_len = len(UI_MESSAGE_HEADER)
-    try:
-        validate_ui_message_header(ui_message)
-    except Exception:
+    if not validate_ui_message_header(ui_message):
         raise AdminError(
             f"Invalid UI attestation message header: {ui_message[:mh_len].hex()}")
 
-    # Extract UD value, UI public key and signer version from message
+    # Extract UI version, UD value, UI public key and signer version from message
+    ui_version = re.match(b"^HSM:UI:(5.[0-9])$", ui_message[:mh_len]).group(1)
     ud_value = ui_message[mh_len:mh_len + UD_VALUE_LENGTH].hex()
     ui_public_key = ui_message[mh_len + UD_VALUE_LENGTH:mh_len + UD_VALUE_LENGTH +
                                PUBKEY_COMPRESSED_LENGTH].hex()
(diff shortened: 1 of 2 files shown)


middleware/admin/verify_attestation.py
```diff
@@ -45,6 +45,26 @@
...
+def validate_signer_message_header(signer_message):
+    minor_offset = len(SIGNER_MESSAGE_HEADER) - 1
+    if signer_message[:minor_offset] != SIGNER_MESSAGE_HEADER[:minor_offset]:
+        raise AdminError()
+    version_minor = signer_message[minor_offset]
+    # The minor version must be a single digit between 0 and 9
+    if version_minor < 48 or version_minor > 57:
+        raise AdminError()
+
+
```
amendelzon, line 67: Why not use simple regular expressions for this? Like for example:

`re.compile(b"^HSM:UI:5.[0-9]$").match(b"HSM:UI:5.3") is not None`

middleware/admin/verify_attestation.py
```diff
@@ -122,7 +142,9 @@ def do_verify_attestation(options):
     ui_message = bytes.fromhex(ui_result[1])
     ui_hash = bytes.fromhex(ui_result[2])
     mh_len = len(UI_MESSAGE_HEADER)
-    if ui_message[:mh_len] != UI_MESSAGE_HEADER:
+    try:
+        validate_ui_message_header(ui_message)
+    except Exception:
```
amendelzon, line 147: Do we need the validation functions to raise an error? Can't we just raise one if the validation function returns `False`?

middleware/admin/verify_attestation.py
```diff
@@ -138,6 +160,7 @@ def do_verify_attestation(options):
                                   mh_len + UD_VALUE_LENGTH + PUBKEY_COMPRESSED_LENGTH +
                                   SIGNER_HASH_LENGTH + SIGNER_ITERATION_LENGTH]
     signer_iteration = int.from_bytes(signer_iteration, byteorder='big', signed=False)
+    ui_version = ui_message[mh_len - 3:mh_len]
```
amendelzon, line 163: If we end up using regular expressions, then we can use a group to gather the version from the header directly without having to use magic numbers, e.g., `re.match(b"^HSM:UI:(5.[0-9])$", "HSM:UI:5.3").group(1)`.

middleware/admin/verify_attestation.py
italo-sampaio: Sure, this sounds good, I'm making that change, thanks.

middleware/admin/verify_attestation.py
italo-sampaio: Sounds good, I'll have these validation functions return a bool instead

middleware/admin/verify_attestation.py
```diff
@@ -23,13 +23,14 @@
...
+import re
 from .misc import info, head, AdminError
 from .utils import is_nonempty_hex_string
 from .certificate import HSMCertificate
 
 
-UI_MESSAGE_HEADER = b"HSM:UI:5.1"
-SIGNER_MESSAGE_HEADER = b"HSM:SIGNER:5.1"
+UI_MESSAGE_HEADER = b"HSM:UI:5.X"
+SIGNER_MESSAGE_HEADER = b"HSM:SIGNER:5.X"
```
amendelzon, line 33: Are these unused now? Can we use them for the regular expressions instead? So that it's easier to update when we need to bump a version.

amendelzon: And also to avoid duplication like we are introducing right now.

italo-sampaio: Pushed a new commit addressing this concern, let me know how it looks.

The solution has one minor drawback though. I had to change the error messages on lines `136` and `180` since now we use the regex match to compute the header length. As a consequence, in the event of an error, the entire message will be printed to the user, not only the header. Not a major problem IMO, since we're not losing any information, but curious to hear your thoughts.

amendelzon (approved):
LGTM, thanks for addressing the comments.

middleware/admin/verify_attestation.py
amendelzon: I understand. Looking at it, I think the resulting error message is still meaningful to somebody trying to debug a problem, so we should be fine.

PR #198, opened 2024-09-10
SGX middleware manager

amendelzon:
For easier reviewing, I've split this into two separate commits:
1st - for general renaming of artifacts into more Ledger-specific names
2nd - for the actual SGX manager implementation


italo-sampaio (approved):
LGTM!

PR #200, opened 2024-09-18
Removing compilation products from repository

amendelzon:


italo-sampaio (approved):
Looks good. Are these all in a `.gitignore` already? Might be worth double-checking just in case.

PR #203, opened 2024-09-30
Added unit tests for the nvmem module

italo-sampaio:
- Added unit tests for `hal/sgx/src/trusted/nvmem.c`
- Some small fixes found by the tests

### DIFF
--- a/firmware/src/hal/include/hal/nvmem.h
+++ b/firmware/src/hal/include/hal/nvmem.h
@@ -32,7 +32,6 @@
 /**
  * @brief Write to non volatile memory
  *
- * @param key The key of the destination in non volatile memory
  * @param dst The destination address in (non volatile) memory
  * @param src The source address to write from
  * @param length The amount of bytes to write
@@ -73,8 +72,10 @@ void nvmem_init();
  * @param key a string key to uniquely identify the block
  * @param addr the base address of the block
  * @param size the size of the block in bytes
+ *
+ * @return whether the block was successfully registered
  */
-void nvmem_register_block(char* key, void* addr, size_t size);
+bool nvmem_register_block(char* key, void* addr, size_t size);
 
 /**
  * @brief Loads registered blocks into memory
--- a/firmware/src/hal/include/hal/nvmem.h
+++ b/firmware/src/hal/include/hal/nvmem.h
@@ -75,7 +75,7 @@ void nvmem_init();
  *
  * @return whether the block was successfully registered
  */
-bool nvmem_register_block(char* key, void* addr, size_t size);
+bool nvmem_register_block(char* key, void* addr, uint8_t size);
 
 /**
  * @brief Loads registered blocks into memory
--- a/firmware/src/hal/include/hal/nvmem.h
+++ b/firmware/src/hal/include/hal/nvmem.h
@@ -75,7 +75,7 @@ void nvmem_init();
  *
  * @return whether the block was successfully registered
  */
-bool nvmem_register_block(char* key, void* addr, uint8_t size);
+bool nvmem_register_block(char* key, void* addr, size_t size);
 
 /**
  * @brief Loads registered blocks into memory
--- a/firmware/src/hal/sgx/test/common/mock.h
+++ b/firmware/src/hal/sgx/test/common/mock.h
@@ -1,0 +1,89 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2021 RSK Labs Ltd
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+// Mock functions for the secret store
+
+/**
+ * @brief Initializes the mock secret store
+ */
+void mock_sest_init();
+
+/**
+ * @brief Mock implementation of sest_exists
+ *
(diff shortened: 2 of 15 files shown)


firmware/src/hal/sgx/test/common/mock.h
amendelzon, line 1: Shouldn't this bear the same name as the `.c` file?

firmware/src/hal/sgx/test/mock/mock_secret_store.c
amendelzon, line 1: Is there a reason we are putting this in the general `mock` directory? Same question goes for the corresponding header file (which, additionally, is in a different directory).

firmware/src/hal/sgx/test/nvmem/Makefile
```diff
@@ -0,0 +1,48 @@
...
+	$(CC) $(COVFLAGS) -o $@ $^
+
+test_nvmem.o: test_nvmem.c
+	$(CC) $(CFLAGS) -c -o $@ $^
+
+nvmem.o: $(SRCDIR)/nvmem.c
+	$(CC) $(CFLAGS) $(COVFLAGS) -c -o $@ $^
+
+mock_secret_store.o: $(MOCKDIR)/mock_secret_store.c
+	$(CC) $(CFLAGS) -c -o $@ $^
```
amendelzon, line 40: We shouldn't need these targets. Check out other unit tests' makefiles to see what I mean.

firmware/src/sgx/src/trusted/system.c
```diff
@@ -206,12 +206,19 @@ bool system_init(unsigned char *msg_buffer, size_t msg_buffer_size) {
...
     nvmem_init();
-    nvmem_register_block("bcstate",
-                         &N_bc_state_var,
-                         sizeof(N_bc_state_var));
-    nvmem_register_block("bcstate_updating",
-                         &N_bc_state_updating_backup_var,
-                         sizeof(N_bc_state_updating_backup_var));
+    if (!nvmem_register_block("bcstate",
+                              &N_bc_state_var,
+                              sizeof(N_bc_state_var))) {
```
amendelzon, line 211: Is there no compilation error with the type change to `uint8_t`?

firmware/src/hal/include/hal/nvmem.h
```diff
@@ -73,8 +72,10 @@ void nvmem_init();
  * @param key a string key to uniquely identify the block
  * @param addr the base address of the block
  * @param size the size of the block in bytes
+ *
+ * @return whether the block was successfully registered
  */
-void nvmem_register_block(char* key, void* addr, size_t size);
+bool nvmem_register_block(char* key, void* addr, uint8_t size);
```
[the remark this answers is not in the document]
italo-sampaio: When we load the blocks from the secret store using `sest_read`, we verify if the secret was correctly read by doing:
```
if (sest_read(key, tmp, nvm_blocks[i].size) == nvm_blocks[i].size) {
    memcpy((uint8_t*)nvm_blocks[i].addr, tmp, nvm_blocks[i].size);
} else {
    LOG("Error loading NVM block <%s>\n", nvm_blocks[i].key);
    clear_blocks();
    free(key);
    return false;
}
```

but `sest_read` returns a `uint8_t`. So, if we allow a block size to be over `255`, this check will never succeed, even if we manage to read it correctly. I see two possible fixes to this: limiting the block size to a `uint8_t` to prevent this inconsistency or having `sest_read` return a `size_t` instead. I'm good with both of them, but just limiting the block size here is the path of least resistance, since the return of `sest_read` is currently treated as a `uint8_t` by the other callers (seed and access) , and our secrets are under this limit. What do you think?

firmware/src/hal/sgx/test/common/mock.h
italo-sampaio: I thought of using a single `mock.h` for all mocks, like we did for the UI and signer tests. So we would have only one `mock.h` header under `common` and individual `mock_{unit}.c`  files `mock`. How does that sound?

firmware/src/hal/sgx/test/mock/mock_secret_store.c
italo-sampaio: See comment above, happy to change it if it doesn't make sense.

firmware/src/hal/sgx/test/nvmem/Makefile
italo-sampaio: Sounds good. I took this pattern from ledger UI and signer tests, will double check others, thanks!

firmware/src/sgx/src/trusted/system.c
italo-sampaio: Nope, no errors or warnings. I don't think `sizeof` is typed is it?

firmware/src/hal/include/hal/nvmem.h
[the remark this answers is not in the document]
amendelzon: I see the problem, but it is a known issue. If we happen to ever register blocks bigger than 255 bytes, then the app will stop working. But the NVM module should allow it, and we should - in the future - fix the secret store (and key value store) to allow for bigger value sizes. It is unclear why the compiler is not warning about not casting on the last parameter of e.g. `sest_read(key, tmp, nvm_blocks[i].size)`. Do you know why?

firmware/src/hal/sgx/test/common/mock.h
amendelzon: Sounds messy. We could have individual `mock_{unit}.{h,c}` and then a single `mock.h` that `#include`s all other headers. That way we get the best of both worlds.

firmware/src/hal/sgx/test/mock/mock_secret_store.c
amendelzon: I see no reason to be guessing locations for files. Keeping both header and source files for everything mock-related under the `mock` directory seems like the most straightforward decision.

firmware/src/hal/sgx/test/common/mock.h
italo-sampaio: Nice, that sounds great, will change to that approach, thanks.

firmware/src/hal/include/hal/nvmem.h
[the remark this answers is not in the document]
italo-sampaio: Got it. In this case, I'll revert to `size_t` and keep the tests limited to 255 bytes with a comment to remind us this is a known limitation. How does that sound?

Regarding the compiler not complaining, I don't know the reason. Will double check the flags we are using, maybe we're not enabling some warnings for the sgx targets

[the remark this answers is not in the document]
amendelzon: Sounds good. Please check the flags and add them if necessary.

firmware/src/sgx/src/trusted/system.c
amendelzon: True, `sizeof` is not typed.

firmware/src/hal/include/hal/nvmem.h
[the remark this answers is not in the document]
italo-sampaio: Reverted back to `size_t`

firmware/src/hal/sgx/test/common/mock.h
italo-sampaio: Fixed

firmware/src/hal/sgx/test/nvmem/Makefile
italo-sampaio: targets removed, added `VPATH` to `common.mk` instead

firmware/src/hal/include/hal/nvmem.h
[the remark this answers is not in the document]
italo-sampaio: Regarding the flags, I double-checked and we are enabling all warnings (and also `-Werror`) for the sgx build. My guess is that the compiler didn't complain because the implicit cast was casting a `uint8_t` to a `size_t`, which is fine, the warning would be present if it was the other way around.

In any case, this reminded me of adding these flags to the tests as well just to make sure we keep the tests warning-free, so I added them and fixed some warnings on `mock_secret_store.c`.

amendelzon (approved):
LGTM!

PR #210, opened 2024-10-18
Unit testing bip32 derivation and hmac_sha512

amendelzon:
- Added bip32 derivation tests on separate source file
- Added hmac_sha512 tests using 3rd party test vectors
- Added test helpers to common x86 test artifacts
- Moved common.mk to common directory
- Updated existing Makefile(s) to account for moved common.mk


italo-sampaio (approved):
LGTM!

PR #211, opened 2024-10-22
Added unit tests for system module

italo-sampaio:


firmware/src/hal/sgx/test/common/common.mk
```diff
@@ -20,17 +20,24 @@
...
+CFLAGS += -iquote $(HALSRCDIR) -iquote $(HALINCDIR)
 CFLAGS += -iquote $(TESTCOMMONDIR)
+CFLAGS += -iquote $(POWHSMSRCDIR)
+CFLAGS += -iquote $(COMMONSRCDIR)
+CFLAGS += -iquote $(SGXSRCDIR)
 CFLAGS += -I$(MOCKDIR)
 CFLAGS += -DHSM_PLATFORM_SGX
 
-VPATH += $(MOCKDIR):$(SRCDIR)
+VPATH += $(MOCKDIR):$(HALSRCDIR):$(TRUSTEDSRCDIR)
```
amendelzon, line 40: What do we need the `HALSRCDIR` for here?

italo-sampaio: That resolves to `firmware/src/hal/sgx/src/trusted`, where all the other hal modules are located (nvmem, secret_store, seed, etc). I just renamed this from `SRCDIR` to `HALSRCDIR` to differentiate from `SGXSRCDIR`, which is where `system.c` is located.

amendelzon: I see, my bad. I've literally just noticed after this comment that you put the unit tests under `firmware/src/hal/sgx/test`, when in reality they belong in (the currently nonexistent) `firmware/src/sgx/test` (since the module you are testing is under `firmware/src/sgx/src`).

italo-sampaio: Sorry, my bad. Yes, it makes sense, there's no point in this being under `hal`, will fix this.

italo-sampaio: Updated to the correct location

amendelzon (approved):
LGTM

PR #220, opened 2024-11-19
Adds MRENCLAVE extractor tool

italo-sampaio:
Adds a middleware utility that extracts the MRENCLAVE value from a signed enclave binary.


amendelzon (changes requested):
We need some unit tests. At least of the core logic classes. Also, would you be able to put all this under an `sgx` module? And potentially split into separate files to have smaller modules. Thanks.

middleware/mrenclave.py
```diff
@@ -0,0 +1,83 @@
...
+# SOFTWARE.
+
+from argparse import ArgumentParser
+from admin.misc import AdminError
+from elftools.elf.elffile import ELFFile
+import json
+from sgx.sgxtypes.sgx_enclave_properties import EnclaveProperties
+
+
+class EnclaveBinary:
```
amendelzon, line 30: This deserves a module of its own I would think. And we don't need to unit test the `main`, just this class should be enough.

italo-sampaio:
We decided to change the approach to a script that calls into `oesign`: https://github.com/rsksmart/rsk-powhsm/pull/221

PR #224, opened 2024-11-25
Fixes coverage report for feature/sgx branch

italo-sampaio:
There were two problems with the coverage report for the feature branch:

1 - The `coverage.yml` workflow was not set to run on pushes to `feature/sgx`, which means that there was no coverage badge available for version `5.3.x`, which was causing the broken coverage badge.

2 - Some unit tests were failing inside the `gen-coverage` script, since internally they try would try to start a new instance of the `hsm:mware` docker container. Since the `gen-coverage` script itself runs in a docker container, these tests were failing. This was introduced basically because these unit tests would fail in the `run-tests` workflow if they were set to run in the host machine. This was fixed by standardizing all the `run-all.sh` scripts for the unit tests not to assume they need to start their docker instance, and move that logic to the `run-test` workflow instead.

NOTE: after that change, an intermittent bug in `srlp` and `btctx` unit tests became more frequent, so the last commit fixes it (details in the commit message).

### DIFF
--- a/firmware/src/hal/sgx/test/run-all.sh
+++ b/firmware/src/hal/sgx/test/run-all.sh
@@ -1,12 +1,21 @@
 #!/bin/bash
-ROOTDIR=$(dirname $0)/../../../../..
-TESTDIR=$(realpath $(dirname $0) --relative-to $ROOTDIR)
-TESTDIRS="nvmem secret_store seed"
-TESTDIRS=${1:-"$TESTDIRS"}
 
-for d in $TESTDIRS; do
-    echo "******************************"
-    echo "Testing $d..."
-    echo "******************************"
-    $ROOTDIR/docker/mware/do-notty-nousb /hsm2/$TESTDIR/$d "make clean test" || exit $?
-done
+if [[ $1 == "exec" ]]; then
+    BASEDIR=$(realpath $(dirname $0))
+    TESTDIRS="nvmem secret_store seed"
+    for d in $TESTDIRS; do
+        echo "******************************"
+        echo "Testing $d..."
+        echo "******************************"
+        cd "$BASEDIR/$d"
+        make clean test || exit $?
+        cd - > /dev/null
+    done
+    exit 0
+else
+    # Script directory
+    REPOROOT=$(realpath $(dirname $0)/../../../../..)
+    SCRIPT=$(realpath $0 --relative-to=$REPOROOT)
+
+    $REPOROOT/docker/mware/do-notty-nousb /hsm2 "./$SCRIPT exec"
+fi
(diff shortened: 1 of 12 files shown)


firmware/src/hal/sgx/test/run-all.sh
amendelzon, line 1: Does this change mean that we can no longer run this locally outside of an `hsm:mware` container?

firmware/src/hal/x86/test/run-all.sh
amendelzon, line 1: Same question as for the other changed `run-all.sh`

firmware/src/hal/sgx/test/run-all.sh
amendelzon: I understand. They don't run locally in my dev environment since I don't have the necessary dependencies installed. And I think that's the spirit of using Docker in our case: forgetting about manually having to resolve dependencies, be that for production or for development. Maybe have something along the lines of what our `lint-c` script does for our `run-all.sh` scripts?

amendelzon: We can just implement the opposite logic: instead of an `exec` argument that executes the tests straight away, we can have the default (no arguments) execute the tests. And a special argument to spin a container and execute. Thoughts?

italo-sampaio: IDK, this just moves the problem somewhere else. If we do this opposite logic, we need to add this new special argument to `run-tests` workflow:

```
(...)
- name: Firmware HAL's x86 unit tests
  run: firmware/src/hal/x86/test/run-all.sh with-docker

- name: Firmware HAL's SGX unit tests
  run: firmware/src/hal/sgx/test/run-all.sh with-docker
  (...)
  ```

Also, the problem remains: if your dev machine doesn't have the requirements, you will still be unable to run the scripts directly, and you will need to memorize these two special cases. That's exactly what we're trying to avoid.

I think the best solution is having all scripts just work straight away on any environment (meaning: the script knows if a docker container is needed, and the user just calls `run-all.sh` without worrying about special arguments) and all special cases are handled internally by the scripts.

Thoughts?

amendelzon: In that case, I'd just unify all the `run-all.sh` to have a `lint-c` like logic (i.e., no arguments spins a container, with `exec` it doesn't). And then add the `exec` argument to all scripts in the coverage generation script (alongside a comment explaining why).

italo-sampaio: Nice, sounds good. I'll make that change, thanks!

amendelzon (approved):
LGTM

PR #225, opened 2024-11-29
SGX endorsement and platform library

amendelzon:
- Added endorsement initialisation and finalisation functions for SGX
- Added system finalisation ecall, used within the untrusted code's finalisation routine
- Removed old stubs for SGX's HAL platform and endorsement
- Added der_utils library to aid with endorsement signature encoding
- Implemented SGX endorsement library
- Implemented SGX platform library
- Adjusting sync acquire lock's function to account for different return types
- Removed now unused hmac_sha256 library from SGX's HAL
- Added SGX's HAL der_utils and endorsement unit tests


italo-sampaio (approved):
LGTM!

PR #226, opened 2024-11-29
Enhance SGX install scripts to install as a systemd service

italo-sampaio:
Builds generated with these changes:

https://github.com/rootstock/powpeg-node-setup/pull/130 (5.3 - SGX platform)
https://github.com/rootstock/powpeg-node-setup/pull/131 (5.3 - Ledger platform)
https://github.com/rootstock/powpeg-node-setup/pull/132 (5.2 - Ledger platform)


amendelzon (approved):
Looks good for now.

We should really not treat installation and service setup as two separate things, but rather as a single operation. Meaning the order of things should be:
- Prompt user for everything needed (except onboarding)
- Install files
- Install service
- Start service
- Onboard
- Restart service
Let's create a task with this as tech debt.

PR #228, opened 2024-12-05
SGX attestation gathering

amendelzon:
- Unified certificates in certificate module
- Moved existing `HSMCertificate` and `HSMCertificateElement` to certificate_v1
- Added new `HSMCertificateV2` and `HSMCertificateV2Element` (and subclasses)
- Added new attestation gathering operation for `adm_sgx`
- Added new attestation gathering module used within `adm_sgx`
- Factored out user-defined value gathering into `admin/misc` module
- Current SGX attestation saving to file is directly translated from value grabbed with `HSM2Dongle`

### DIFF
--- a/middleware/admin/certificate_v1.py
+++ b/middleware/admin/certificate_v1.py
@@ -1,0 +1,251 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2021 RSK Labs Ltd
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
+# use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+# of the Software, and to permit persons to whom the Software is furnished to do
+# so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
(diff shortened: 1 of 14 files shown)


middleware/admin/certificate_v1.py
```diff
@@ -0,0 +1,251 @@
...
+            return verifier_pubkey.ecdsa_verify(
+                message, verifier_pubkey.ecdsa_deserialize(bytes.fromhex(self.signature)))
+        except Exception:
+            return False
+
+    def get_value(self):
+        return self.EXTRACTORS[self.name](bytes.fromhex(self.message)).hex()
+
+
+class HSMCertificate:
```
italo-sampaio, line 124: Is there a reason for not moving `HSMCertificate` class to `certificate.py` and keeping the common logic between v1 and v2 there instead?

Then this file would define `HSMCertificateV1` with v1 specific logic, similarly to what is done in `certificate_v2.py`.

italo-sampaio (approved):
LGTM! Left a comment regarding the relationship between certificate_v1 / certificate_v2, but take that more as a question than a suggestion.

middleware/admin/certificate_v1.py
amendelzon: Didn't think much about it to be honest. I guess just to not have an almost empty `HSMCertificate` class with a single class method. Something to improve in the future. Thanks.

PR #231, opened 2024-12-20
Certificate v2 validator

amendelzon:
- Replacing plain public key in validate_and_get_values method for a root of trust object with get_pubkey method
- Using inherited validate_and_get_values in HSMCertificateV2
- Added HSMCertificateRoot to certificate version 1 to act as root of trust in existing validations
- Updated verify ledger attestation accordingly


italo-sampaio (approved):
LGTM!

PR #233, opened 2024-12-27
OSSF Scorecard

bcodesido:
- Implement OSSF Scorecard.
- Implement required tools like:
  - Dependency review
  - Dependabot
- Pin workflows actions
- Pin docker image's versions
- Set job level worflow permissions


amendelzon (approved):
LGTM

PR #240, opened 2025-01-06
Certificate V2 x509_pem type element validation

amendelzon:
- Implemented `HSMCertificateV2ElementX509`'s `is_valid` method
- Added root of trust element validation in SGX attestation verification

### DIFF
--- a/middleware/admin/certificate_v2.py
+++ b/middleware/admin/certificate_v2.py
@@ -259,7 +259,7 @@ class HSMCertificateV2ElementX509(HSMCertificateV2Element):
             # IMPORTANT: for now, we only allow verifying the validity of an
             # HSMCertificateV2ElementX509 using another HSMCertificateV2ElementX509
             # instance as certifier. That way, we simplify the validation procedure
-            # and ensure maximum compatibility wrt the underlying library used
+            # and ensure maximum use of the underlying library's capabilities
             # (cryptography)
             if not isinstance(certifier, type(self)):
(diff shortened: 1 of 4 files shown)


middleware/admin/certificate_v2.py
```diff
@@ -254,13 +255,40 @@ def certificate(self):
         return self._certificate
 
     def is_valid(self, certifier):
-        return True
+        try:
+            # IMPORTANT: for now, we only allow verifying the validity of an
+            # HSMCertificateV2ElementX509 using another HSMCertificateV2ElementX509
+            # instance as certifier. That way, we simplify the validation procedure
+            # and ensure maximum compatibility wrt the underlying library used
```
italo-sampaio, line 262: Typo? wrt -> with?

italo-sampaio (approved):
LGTM!

middleware/admin/certificate_v2.py
amendelzon: wrt = with respect to

you reckon I should I use the long version?

italo-sampaio: Up to you, really. This confused me, so using the long version would probably not hurt, but the message is clear anyway.

amendelzon: Ended up rephrasing it cause the word "compatibility" didn't actually reflect what I was trying to say. Let me know wdyt.

PR #242, opened 2025-01-07
SGX onboarding attestation gathering and verification

amendelzon:
- Updated setup script to gather and verify attestation after onboarding is completed
- Added runtime attestation dependencies to the SGX distribution `Dockerfile`


italo-sampaio (approved):
LGTM!

PR #245, opened 2025-01-10
Fixes signature for finalise function

italo-sampaio:
Fixes the signature of finalise function so that it conforms with the expected signal handler function signature


italo-sampaio:
will create a new PR targeting the correct branch

PR #259, opened 2025-01-14
SGX heartbeat documentation

amendelzon:
- Added note to heartbeat documentation stating that it is currently unsupported for SGX
- Added corresponding unsupported notes in protocol operations within the protocol documentation


italo-sampaio (approved):
LGTM!

PR #262, opened 2025-01-16
Adds unsigned enclave binary to distribution

italo-sampaio:
The hash of the unsigned enclave will be included in the reproducible builds information, so it also needs to be part of the distribution


amendelzon (commented):
Depending on what happens with [the automated build changes PR](https://github.com/rootstock/powpeg-automated-build/pull/75), we might not need this at all.

italo-sampaio:
Closing as this won't be necessary

PR #267, opened 2025-01-20
Added APDU buffer pointer validation to SGX enclave init sequence

amendelzon:
- Using `oe_is_outside_enclave` to validate the APDU buffer in `system_init`
- Added unit test case


italo-sampaio (approved):
LGTM!

PR #268, opened 2025-01-21
Moves finalise logic out of signal handler

italo-sampaio:
Signal handler now only sets a flag that is checked in main.

### DIFF
--- a/firmware/src/sgx/src/untrusted/main.c
+++ b/firmware/src/sgx/src/untrusted/main.c
@@ -54,6 +54,9 @@ struct arguments {
     char *enclave_path;
 };
 
+// Global flag to indicate that the application should stop
+static sig_atomic_t G_stop_requested = 0;
+
 // Argp individual option parsing function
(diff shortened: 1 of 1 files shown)


firmware/src/sgx/src/untrusted/main.c
```diff
@@ -157,6 +162,11 @@ int main(int argc, char **argv) {
     unsigned int tx = 0;
 
     while (true) {
+        if (G_stop_requested) {
+            G_stop_requested = 0;
+            finalise_with(0);
+        }
+
```
amendelzon, line 169: I would just exit the loop instead, and call `finalise_with(0);` instead of the unexpected loop exit log. No need to reset the flag either.

italo-sampaio: Pushed a new commit, let me know how it looks

amendelzon (approved):
LGTM

PR #276, opened 2025-01-22
Max APDU buffer size checks

amendelzon:
- Renaming empty buffer error to more generic invalid buffer error
- Added max buffer size checks on powHSM, Ledger UI bootloader and Ledger UI heartbeat main I/O logic
- Removed lower untrusted SGX & TCPSigner I/O layer disconnection on inconsistent request
- Added request buffer emptying after an inconsistent request to untrusted SGX & TCPSigner I/O layer
- Added new RawCommand test type to firmware test framework
- Added test cases for empty APDU and oversized APDU to firmware tests
- Added test cases for Ledger UI bootloader and Ledger UI heartbeat


italo-sampaio (approved):
LGTM!

PR #280, opened 2025-01-23
Local copy of APDU buffer for processing

amendelzon:
System module now keeps an enclave-only copy of the APDU buffer and performs a copy from the host APDU buffer before processing and a copy back to the host APDU buffer after processing is done. In this way we prevent double buffer fetching.

### DIFF
--- a/firmware/src/sgx/src/trusted/system.c
+++ b/firmware/src/sgx/src/trusted/system.c
@@ -170,8 +170,6 @@ unsigned int system_process_apdu(unsigned int rx) {
 }
 
 bool system_init(unsigned char* msg_buffer, size_t msg_buffer_size) {
-    // Setup the host APDU buffer
-
     // Validate that host and enclave APDU buffers have the same size
     if (msg_buffer_size != sizeof(apdu_buffer)) {
         LOG("Expected APDU buffer size to be %lu but got %lu\n",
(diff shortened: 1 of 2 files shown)


firmware/src/sgx/src/trusted/system.c
```diff
@@ -160,27 +161,34 @@ static external_processor_result_t system_do_process_apdu(unsigned int rx) {
...
+    memcpy(host_apdu_buffer, apdu_buffer, sizeof(apdu_buffer));
+    return tx;
 }
 
 bool system_init(unsigned char* msg_buffer, size_t msg_buffer_size) {
-    // Setup the shared APDU buffer
-    if (msg_buffer_size != EXPECTED_APDU_BUFFER_SIZE) {
-        LOG("Expected APDU buffer size to be %u but got %lu\n",
-            EXPECTED_APDU_BUFFER_SIZE,
+    // Setup the host APDU buffer
```
italo-sampaio, line 173: should we remove this comment?

italo-sampaio (approved):
Looking good! I just think there's an extra comment that should be removed

firmware/src/sgx/src/trusted/system.c
amendelzon: done, thanks.

italo-sampaio (approved):
LGTM!

PR #282, opened 2025-01-24
Fixes sgx main loop not handling signals properly

italo-sampaio:
Main loop blocks while waiting for client data. For this reason, we can't just check for the stop condition synchronously. In any case, we still need to ensure that the finalise logic is only called once.

This PR reverts commit 1ec2d3553c7cc3ce9ac31f70263c4926d1c343dd and introduces a global counter to ensure `finalise_with` is only called once.


amendelzon (approved):
LGTM

PR #283, opened 2025-01-24
Common modules linking approach change

amendelzon:
- All common modules for the HAL now lie in `firmware/src/hal/common_linked`, and are not included in `Makefile`s, but symlinked from necessary source directories (e.g., `firmware/src/x86`).
- Moved existing unit tests
- Renamed directory in CI workflow
- Incidentally fixed compiler warning in UI unit test


italo-sampaio (approved):
LGTM!

PR #285, opened 2025-01-24
Fixed firmware test suite for SGX

amendelzon:
- Tests cases can now specify that they run either on `ledger`, `sgx`, `tcpsigner`, `all` or any combination
- Heartbeat test command and dongle operations now account for failing cases and exception code expectations
- Updated `runOn` value for test cases
- Added failing heartbeat test case for SGX only
- Updated unit tests


italo-sampaio (approved):
LGTM!

PR #290, opened 2025-01-30
Moved bip32_parse_path to x86 only

amendelzon:
- Split bip32 module into bip32 (for derivation) and bip32_path (for path parsing)
- Split existing unit tests


italo-sampaio (approved):
LGTM!

PR #292, opened 2025-02-03
Adds header checking to secret store

italo-sampaio:
Introduces a header to the data to be sealed by the secret store.

Now a header is added to the secret before sending it to the seal API.
The header contains a type constant that is verified when the secret
is retrieved. This prevents the files containing the sealed blobs from
being swapped and treated as valid data by the system.

### DIFF
--- a/firmware/src/hal/sgx/src/trusted/secret_store.c
+++ b/firmware/src/hal/sgx/src/trusted/secret_store.c
@@ -32,6 +32,7 @@
 
 #include "hal/platform.h"
 #include "hal/log.h"
+#include "hal/constants.h"
 #include "sha256.h"
 #include "secret_store.h"
 #include "hsm_t.h"
@@ -49,28 +50,14 @@
 
 #define MAX_BLOB_SIZE (1024 * 1024)
 
-#define SECRET_PAYLOAD_SIZE(secret_size) (secret_size - sizeof(secret_header_t))
-#define SEST_HASH_SIZE (32)
-
-// To avoid performing dynamic memory allocation whenever we need to read a
-// secret, we use this static buffer to temporarily store the sealed blobs we
-// read from the untrusted modules.
-static uint8_t G_secrets_buffer[MAX_BLOB_SIZE];
+// To avoid performing dynamic memory allocation whenever we need to manipulate
+// data, we use these static buffers to store the sealed and unsealed data,
+// respectively.
+static uint8_t G_sealed_buffer[MAX_BLOB_SIZE];
+static uint8_t G_unsealed_buffer[MAX_BLOB_SIZE];
 
 // The sha256 context used for hash operations.
 static SHA256_CTX G_sha256_ctx;
-
-// The blob format of the sealed secrets we send to the seal API.
-// This includes the payload and a sha256 hash of the key for validation
-// purposes.
-typedef struct {
-    uint8_t key_hash[SEST_HASH_SIZE];
-} secret_header_t;
-
-typedef struct {
-    secret_header_t header;
-    uint8_t payload[0];
-} secret_t;
 
 // The file format of the sealed secrets. This is the format we use to read and
 // write sealed the data to disk.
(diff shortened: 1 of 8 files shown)


firmware/src/hal/sgx/src/trusted/secret_store.c
```diff
@@ -48,18 +49,64 @@
 
 #define MAX_BLOB_SIZE (1024 * 1024)
 
+#define SECRET_PAYLOAD_SIZE(secret_size) (secret_size - sizeof(secret_header_t))
+#define SEST_HASH_SIZE (32)
```
amendelzon, line 53: Can we use `HASH_LENGTH` from `hal/constants.h` instead so we avoid duplicating constants?

firmware/src/hal/sgx/src/trusted/secret_store.c
```diff
@@ -48,18 +49,64 @@
...
+// purposes.
+typedef struct {
+    uint8_t key_hash[SEST_HASH_SIZE];
+} secret_header_t;
+
+typedef struct {
+    secret_header_t header;
+    uint8_t payload[0];
+} secret_t;
+
```
amendelzon, line 74: I'm a bit concerned about alignment here. Can we guarantee this is held in memory as contiguous bytes?

firmware/src/hal/sgx/src/trusted/secret_store.c
```diff
@@ -48,18 +49,64 @@
...
+/**
+ * @brief Checks if the given secret header is valid for the given key.
+ *
+ * @param key The key to validate the header against.
+ * @param header The header to validate.
+ *
+ * @returns true if the header is valid, false otherwise.
+ */
+static bool check_secret_header(const char* key,
```
amendelzon, line 102: Please rename to `is_secret_header_valid` to improve readability.

firmware/src/hal/sgx/src/trusted/secret_store.c
```diff
@@ -48,18 +49,64 @@
 
 #define MAX_BLOB_SIZE (1024 * 1024)
 
+#define SECRET_PAYLOAD_SIZE(secret_size) (secret_size - sizeof(secret_header_t))
+#define SEST_HASH_SIZE (32)
+
 // To avoid performing dynamic memory allocation whenever we need to read a
 // secret, we use this static buffer to temporarily store the sealed blobs we
 // read from the untrusted modules.
 static uint8_t G_secrets_buffer[MAX_BLOB_SIZE];
```
amendelzon, line 58: Shoudn't we allow additional space for the header here when we are using this for writing?

firmware/src/hal/sgx/test/mock/mock_seal.c
```diff
@@ -128,8 +128,8 @@ void assert_oe_seal_called_with(const void* plugin_id,
                                 size_t additional_data_size) {
     assert(G_oe_seal_args.plugin_id == plugin_id &&
```
amendelzon, line 129: I've just realised the existence of this assert of multiple conditions joined by `&&`. We should rewrite this as multiple `assert`s to aid in debugging.

firmware/src/hal/sgx/test/secret_store/test_secret_store.c
```diff
@@ -40,6 +40,31 @@
...
+    // sha256 hash of the key
+    uint8_t* unsealed_secret;
+    // The size of the unsealed secret
+    size_t unsealed_size;
+    // The sealed secret, which is the unsealed secret prepended with the string
+    // "SEALED - "
+    uint8_t* sealed_secret;
+    // The size of the sealed secret
+    size_t sealed_size;
+} secret_t;
```
amendelzon, line 67: I find the name of this mocking structure very confusing due to the existence of `secret_t` in the secret store implementation. Can we give this a different and more meaningful name please?

amendelzon (changes requested):
Looking alright, a bit concerned about the amount of complexity in data type definitions on something that's essentially adding a 32-byte header to a string of bytes. Specifically, I'm worried about the struct definitions not behaving exactly as we intend them to bytewise, and also about the mixed use of dynamic memory allocation and static blobs. We should definitely try to find an easier, more straightforward implementation.

firmware/src/hal/sgx/src/trusted/secret_store.c
italo-sampaio: sounds good

firmware/src/hal/sgx/src/trusted/secret_store.c
italo-sampaio: That's a good question. The `blob` is still composed of the header + payload, so I think this line specifically is correct. But since we increased the size of the data to be sealed, you're probably right that in theory, we should probably increase `MAX_BLOB_SIZE`.

Looking at real data, though, it seems like this 1MB limit is already more than enough to our practical needs. What do you think?

firmware/src/hal/sgx/test/secret_store/test_secret_store.c
italo-sampaio: Sounds good, I'll rename it so that it makes its meaning clearer

firmware/src/hal/sgx/src/trusted/secret_store.c
```diff
@@ -60,6 +66,56 @@ typedef struct {
...
+ * @param key The key to validate the header against.
+ * @param data The buffer containing the data to validate.
+ *
+ * @returns true if the header is valid, false otherwise.
+ */
+static bool is_header_valid(const char* key, const uint8_t* data) {
+    uint8_t expected_header[HASH_LENGTH];
+    add_header(key, NULL, 0, expected_header, HASH_LENGTH);
+
+    return memcmp(expected_header, data, HASH_LENGTH) == 0;
```
amendelzon, line 116: We need to make sure the comparison doesn't read out of bounds.

amendelzon (changes requested):
Looks good, made a small but relevant comment we need to address.

firmware/src/hal/sgx/src/trusted/secret_store.c
italo-sampaio: Done! See last commit

amendelzon (approved):
LGTM

PR #295, opened 2025-02-04
Moved bip32 and dependent libraries to SGX HAL

amendelzon:
- Moved `bip32`, `hmac_sha512` and `endian` libraries to SGX HAL
- Moved corresponding unit tests to SGX HAL
- Incidentally adding missing `common.mk` headers
- Incidentally removing unused `common.mk`


italo-sampaio (approved):
LGTM!

PR #297, opened 2025-02-05
Increases pip install socket timeout on mware Dockerfile

italo-sampaio:
Increases the socket timeout of the pip install command from 15s to 60s. This helps prevent CI workflow failures due to temporary network instability.


amendelzon (approved):
LGTM

PR #299, opened 2025-02-05
Unifying pin policy

amendelzon:
- Added `pin_policy` library implementing both Ledger and SGX pin policies
- Updated SGX's trusted HAL access module to use the pin policy library for password validation
- Updated the Ledger UI to use the pin policy library for pin validation
- Added `pin_policy` unit tests
- Updated failing unit tests


italo-sampaio (approved):
LGTM!

PR #304, opened 2025-02-11
Show signer version on Ledger screen

italo-sampaio:
Shows the currently running version on the Ledger screen during signer
execution


amendelzon (approved):
LGTM!

PR #306, opened 2025-02-14
Enhance Ledger setup script to explicitly remove Signer and UI apps

italo-sampaio:
This addresses an issue where the setup script would fail if the user has the signer and UI app already installed. To address this, we now explicitly delete both apps if they exist before proceeding with the onboard.


amendelzon (approved):
LGTM

PR #307, opened 2025-02-17
Updates ubuntu image on GH workflows

italo-sampaio:
GitHub will deprecate ubuntu-20.04 runners in April 2025. This commit updates the target image from all workflows to use ubuntu-latest instead.

Incidentally, removes `feature/sgx` target from the coverage workflow, since this branch is not used anymore


amendelzon (approved):
LGTM

PR #310, opened 2025-02-17
Adds development branches as targets for codeQL and coverage workflows

italo-sampaio:
- Adds develop-5.2 and develop-5.4 as targets for the codeQL and coverage workflows
- Renames remote destination for coverage reports to avoid conflicts with development branches


amendelzon (approved):
LGTM

PR #318, opened 2025-02-26
Automatically delete all pre-installed apps on Ledger before the onboard

italo-sampaio:
- Expose the `listApps` command on `lbutils`
- Adds a `post_process` optional function that parses the raw output of ledgerblue utility into a format we can easily use
- Enhances the `setup` script so that it automatically checks and deletes all pre-installed apps before proceeding with the onboard
- Adds unit tests to the new `listApps` utility
- Updates distribution documentation

### DIFF
--- a/dist/ledger/README-cli.md
+++ b/dist/ledger/README-cli.md
@@ -40,12 +40,11 @@ Press [Enter] to continue
 
 Pressing `Enter` will proceed with the installation. Pay attention to the instructions both on the terminal and on the device's screen, since a few confirmations are required.
 ```
-Removing the Bitcoin App...
+Removing the existing installed apps (if any)...
 The Ledger will prompt for 'Allow Unknown Manager'. Please accept it.
+Removing the UX App...
 If the Ledger prompts for 'Remove app' followed by the app name and identifier, then please accept it.
-Removing the Ethereum App...
-If the Ledger prompts for 'Remove app' followed by the app name and identifier, then please accept it.
-Removing the Fido App...
+Removing the RSK Sign App...
 If the Ledger prompts for 'Remove app' followed by the app name and identifier, then please accept it.
 Removing the existing certification authority (if any)...
 If the Ledger prompts for 'Revoke certificate' followed by the certificate name and its public key, then please accept it.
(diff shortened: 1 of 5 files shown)


middleware/lbutils.py
```diff
@@ -20,18 +20,40 @@
...
+        line = line.strip()
+        if line.startswith("[{") and line.endswith("}]"):
+            data = ast.literal_eval(line)
+            if isinstance(data, list) and all(
+                    isinstance(app_dict, dict) for app_dict in data):
+                for app_dict in data:
+                    app_list.append(app_dict["name"])
+                break
+    if app_list:
+        return "\n".join([f"Installed app: {app}" for app
```
amendelzon, line 41: What does the prefix `Installed app: ` serve the purpose of?

dist/ledger/README-cli.md
```diff
@@ -40,12 +40,11 @@ Press [Enter] to continue
 
 Pressing `Enter` will proceed with the installation. Pay attention to the instructions both on the terminal and on the device's screen, since a few confirmations are required.
 ```
-Removing the Bitcoin App...
+Removing the existing installed apps (if any)...
 The Ledger will prompt for 'Allow Unknown Manager'. Please accept it.
+Removing the UX App...
```
amendelzon, line 45: Is this still in the actual output?

amendelzon (commented):
Looking good, just a couple of questions.

middleware/lbutils.py
italo-sampaio: That exists so we can `grep` only the lines of interest in the `setup` script. For example, if the device has no apps installed, a `No apps to remove.` message is displayed (and that's not an error), and we need to make a distinction between a valid list of apps and any other output.

Also, thinking on the use case of `lbutils` as a standalone tool, not only as part of the onboard routine, it felt a bit awkward to just output the list of apps (or nothing if no apps are installed) with no additional context.

What do you think?

dist/ledger/README-cli.md
italo-sampaio: Yes, we do `echo -e "\e[1;32mRemoving the $app App...\e[0m"` in `setup` script, this was copied directly from the script output.

middleware/lbutils.py
amendelzon: Good point. Two things:
1. One way of addressing the need for grepping is just removing any useless output in the preprocessing.
2. The awkwardness I guess depends on what the user expects. If you're using a "list apps" command-line utility, and it literally just outputs a list of names, then I don't see a problem (this is actually the behavior of many UNIX-like command-line utilities).
All of this said, up to you.

dist/ledger/README-cli.md
amendelzon: Oh, gotcha. Thanks for clarifying.

amendelzon (approved):
LGTM!

PR #319, opened 2025-03-03
Fix DER encoding buffer overflow

amendelzon:
- Fix for potential buffer overflow
- Covered with test case


italo-sampaio (approved):
LGTM!

PR #321, opened 2025-03-06
Enhances Ledger setup and upgrade script to use `tput` for specific colors in user-facing messages

italo-sampaio:
- Leverages `tput` utility to set message colors instead of using ANSI escape codes explicitly
- Adds fallback logic so that the script doesn't attempt to set colors if the terminal doesn't support this feature

See details on each individual commit message


amendelzon (approved):
LGTM

italo-sampaio:
Rebased to `master`

PR #323, opened 2025-03-07
Move oeedger8r generated sources to their own directory

italo-sampaio:
These sources are not versioned, and are generated during the build. Having them in the same directory as the other source files was causing an inconsistency on the reproducible builds, since the `TRUSTED_SRC` / `UNTRUSTED_SRC` variables would expand to different values depending on whether or not a build had already been performed.

This is why running a `make clean` followed by a `make build` would result in a different digest / mrenclave pair compared to running `make clean build` on a directory that already contained these sources.

### DIFF
--- a/firmware/src/sgx/.gitignore
+++ b/firmware/src/sgx/.gitignore
@@ -2,10 +2,10 @@
 private.pem
 
 # OE generated files
-src/trusted/hsm_args.h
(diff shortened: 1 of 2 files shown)


italo-sampaio:
Updated to target the correct branch

amendelzon (approved):
LGTM

PR #324, opened 2025-03-07
Ignoring SGX tool-generated files for C linting

amendelzon:


italo-sampaio (approved):
Nice catch, thanks for that!

PR #331, opened 2025-03-17
Use mbedtls for hmac_sha512 and non-midstate sha256 implementation on SGX HAL 

italo-sampaio:
Replaces the custom implementation of HMAC-SHA512 and SHA256 (non-midstate) with the one provided by mbedtls.

### DIFF
--- a/firmware/src/hal/sgx/src/trusted/hmac_sha512.c
+++ b/firmware/src/hal/sgx/src/trusted/hmac_sha512.c
@@ -22,376 +22,50 @@
  * IN THE SOFTWARE.
  */
 
-/**
- * Taken from https://github.com/someone42/hardware-bitcoin-wallet @
- * 102c300d994712484c3c028b215f90a6f99d6155 and adapted for use with
- * the powHSM HAL by RootstockLabs. LICENSE transcribed below.
- */
-
-/*
-  Copyright (c) 2011-2012 someone42
-  All rights reserved.
-
-  Redistribution and use in source and binary forms, with or without
-  modification, are permitted provided that the following conditions are met:
-
-      Redistributions of source code must retain the above copyright notice,
-      this list of conditions and the following disclaimer.
-
-      Redistributions in binary form must reproduce the above copyright notice,
-      this list of conditions and the following disclaimer in the documentation
-      and/or other materials provided with the distribution.
-
-  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
-  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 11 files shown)


firmware/src/hal/sgx/src/trusted/hmac_sha512.h
```diff
@@ -68,10 +68,23 @@
...
+ * @param out A byte array where the HMAC-SHA512 hash value will be written.
+ *            This must have space for SHA512_HASH_LENGTH bytes.
+ * @param key A byte array containing the key to use in the HMAC-SHA512
+ *            calculation. The key can be of any length.
+ * @param key_length The length, in bytes, of the key.
+ * @param text A byte array containing the message to use in the HMAC-SHA512
+ *             calculation. The message can be of any length.
+ * @param text_length The length, in bytes, of the message.
+ */
+void hmac_sha512(uint8_t *out,
```
amendelzon, line 84: Probably a good time to change the return value to `bool` since there are paths that can fail with this implementation. We'd also need to account for this in `bip32.c`.

firmware/src/hal/sgx/src/trusted/hmac_sha512.c
```diff
@@ -22,376 +22,50 @@
...
+    if (!mbedtls_md_info_from_type(MBEDTLS_MD_SHA512)) {
+        LOG("Error: SHA-512 not supported\n");
+        return;
     }
-    // Calculate hash = H((K_0 XOR ipad) || text).
-    sha512Begin(&hs64);
-    for (i = 0; i < sizeof(padded_key); i++) {
-        sha512WriteByte(&hs64, (uint8_t)(padded_key[i] ^ 0x36));
+
+    md_info = mbedtls_md_info_from_type(MBEDTLS_MD_SHA512);
```
amendelzon, line 46: Can't we do this before the if and use `md_info` for the failure condition?

firmware/src/hal/sgx/src/trusted/hmac_sha512.c
```diff
@@ -22,376 +22,50 @@
...
-        }
-        sha512Finish(padded_key, &hs64);
+    mbedtls_md_context_t ctx;
+    int use_hmac;
+    const mbedtls_md_info_t *md_info;
+
+    mbedtls_md_init(&ctx);
+    if (!mbedtls_md_info_from_type(MBEDTLS_MD_SHA512)) {
+        LOG("Error: SHA-512 not supported\n");
+        return;
```
amendelzon, line 43: `goto hmac_sha512_error;`?

amendelzon (changes requested):
Looks good, a return type I think it'd be good to address and a couple of minor questions.

Also a question I couldn't post on the file itself (`hal/sgx/src/trusted/hash.c`, line `31`): don't we need to free the context there too if init fails?

italo-sampaio:
Tackled all suggestions, thanks for the review!

amendelzon (approved):
LGTM

PR #334, opened 2025-03-18
Fixes misleading log message

italo-sampaio:
Ensure that server messages were successfully  sent to the client before logging the event.

### DIFF
--- a/middleware/comm/server.py
+++ b/middleware/comm/server.py
@@ -87,15 +87,13 @@ class _RequestHandler:
                 self.logger.info("=> [%s]: %s", client_address, output)
 
     def _reply(self, wfile, output):
-        success = False
         try:
             wfile.write(output.encode(self.ENCODING))
             wfile.write("\n".encode(self.ENCODING))
-            success = True
+            return True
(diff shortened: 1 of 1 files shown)


middleware/comm/server.py
```diff
@@ -81,15 +82,20 @@ def handle(self, client_address, rfile, wfile):
...
+            success = self._reply(wfile, output)
+            if success:
+                self.logger.info("=> [%s]: %s", client_address, output)
 
     def _reply(self, wfile, output):
+        success = False
         try:
             wfile.write(output.encode(self.ENCODING))
             wfile.write("\n".encode(self.ENCODING))
+            success = True
```
amendelzon, line 94: This is reporting a code scanning issue. Can we just `return true` at the end of the `try` block and `return false` after the logging on the `except` block? Wouldn't that achieve the same without the need for a variable?

italo-sampaio: Thanks for the suggestion, this is fixed now

amendelzon (approved):
LGTM

PR #335, opened 2025-03-19
SGX upgrade workflow

amendelzon:
- Added new instruction on trusted system module for upgrades
- Added upgrade module that handles upgrade instruction (export & import)
- Added upgrade module initialisation upon system initialisation
- Basic mock upgrade workflow implementation
- Added helper REQUIRE_NOT_ONBOARDED macro
- Added upgrade operation init and APDU processing unit tests on system unit tests
- Added upgrade module unit tests
- Fixed failing unit tests


italo-sampaio (approved):
LGTM!

PR #339, opened 2025-03-26
Improvements on SGX setup scripts

italo-sampaio:
- `install_service` now installs all the assets, creates and initializes the systemd service
- `setup` script now interacts with the installed systemd service instead of calling the binary powhsm directly
- Network name is now selected automatically without user intervention
- Renamed service to `powhsmsgx`
- Renamed network basename to  `powhsmsgx_net`
- Installation aborts if a systemd service with the same name already exists


dist/sgx/hsm/start
```diff
@@ -19,11 +19,13 @@ echo
...
 DOCKER_USER="$(id -u):$(id -g)"
 HOSTNAME="SGX"
-NETWORK=${NETWORK:-net_sgx}
+NETWORK=${NETWORK:-powhsmsgx_net}
 PORT=7777
 DOCKER_PORT="$PORT:$PORT"
+SGX_PRV_GID=$(getent group sgx_prv | cut -d: -f3)
 
 docker run --rm --name $DOCKER_CNT --user $DOCKER_USER -v $WORKDIR:/hsm \
+    --group-add $SGX_PRV_GID \
```
amendelzon, line 28: What is this needed for?

dist/sgx/scripts/install_service
```diff
@@ -1,67 +1,197 @@
...
+    error "Could not copy the .env file to the installation directory."
+}
+
+function startPowHsm() {
+    systemctl start $SERVICE_NAME
+	error "Failed to start the service."
+}
+
+function createHsmUser() {
+    if ! id -u hsm >/dev/null 2>&1; then
```
amendelzon, line 117: We should probably rename the user to `powhsm`. Also, do you think it would make sense aborting if the user already exists? Why would that username exist?

dist/sgx/scripts/setup
```diff
@@ -4,28 +4,16 @@ pushd $(dirname $0)/.. > /dev/null
...
-
-function expandBinaries() {
-    cleanBinaries
+ADMIN_BIN="$ADMIN_DIR/adm_sgx"
 
-    mkdir -p $ADMIN_DIR
-    tar -xzmf $ADMIN_BUNDLE -C $ADMIN_DIR
-}
+# The admin command used to interact with the powhsmsgx service
+ADMIN_CMD="$ADMIN_BIN --host SGX"
```
amendelzon, line 16: Shall we rename the host to `powhsmsgx`? I feel like `SGX` is both too generic and also uppercase which is not ideal.

dist/sgx/scripts/setup
```diff
@@ -36,29 +24,23 @@ PUBLIC_KEY_FILE="$EXPORT_DIR/public-keys.txt"
...
+    cleanBinaries
+
+    mkdir -p $ADMIN_DIR
+    tar -xzmf $ADMIN_BUNDLE -C $ADMIN_DIR
 }
 
 function error() {
     if [ $? -ne 0 ]; then
-        print_error "Error comunicating with the dongle. Please check connection and restart the process."
+        print_error "Error comunicating with the pwhsm service. Please restart the process."
```
amendelzon, line 43: `powhsm`

dist/sgx/hsm/start
italo-sampaio: I added in the commit message:

> The container needs to be part of the sgx_prv group, otherwise it won't be
> able to read /dev/sgx_provision
> 
> NOTE: the reason why this used to work without setting the group was that
> we would perform the attestation step of the onboarding while running the
> powhsm binary directly from the host machine with sudo. After changing the
> setup to perform these steps against the running service (which runs inside a
> docker container), we now need that container to be part of the `sgx_prv`
> group.

Also, see [this document by intel](https://builders.intel.com/docs/networkbuilders/intel-software-guard-extensions-intel-sgx-key-management-reference-application-kmra-on-intel-xeon-scalable-processors-user-guide-1698412302.pdf), session 5.12.3 - Enclave not authorized to run in task.

This is the error I mentioned yesterday, we were hitting exactly this error: `[error_driver2api sgx_enclave_common.cpp:247] Enclave not authorized to run, .e.g. provisioning enclave hosted in app without access rights to /dev/sgx_provision. You need add the user id to group sgx_prv or run the app as root.` when trying to do the attestation.

amendelzon (changes requested):
Looking good. A few comments and questions, mainly regarding naming.

dist/sgx/hsm/start
amendelzon: ok, gotcha. this is good.

dist/sgx/scripts/setup
italo-sampaio: Sounds good, I can do that change

dist/sgx/scripts/install_service
italo-sampaio: Renaming the user sounds good.

Reg. aborting, the use case I'm thinking of is the re-onboarding. If, for some reason, the user needs to use these scripts in a machine that already hosts the service, we will need a checklist of what needs to be removed. So far, we only have two items: stop / remove the service and now the user. But even then, it starts to sound like we need an `uninstall` script that performs these steps in case the user knows what he's doing, what do you think?

Maybe that's a separate task. IDK. In any case I'll rename to `powhsm` as you suggested

amendelzon: Yeah, sounds like we probably need an `uninstall` script of some sort. Let's abort if the user exists (if the use case is a re-onboarding, then the service will probably also exist and the installation would abort before reaching the user creation section anyway).

amendelzon (dismissed):
LGTM

dist/sgx/setup-new-powhsm
```diff
@@ -3,21 +3,37 @@
...
 
+print_info "Welcome to the SGX powHSM Setup for RSK"
+echo
+
+$ROOT_DIR/scripts/install_service
+if [ $? -ne 0 ]; then
+	print_error "Error during the powhsm service installation, aborting"
+	exit 1
+fi
+
```
amendelzon, line 22: Why is there a second argument passed in to the `run_with_docker` script?

amendelzon (approved):
LGTM, left a question.

dist/sgx/setup-new-powhsm
italo-sampaio: It's the optional `-v` flag. This logic didn't change, if we pass a `-v`, `docker build` runs without the `-q` flag:

```
if [[ "$2" != "-v" ]]; then
    QUIET="-q"
    echo -e "Run with '-v' if you want to see progress detail\e[94m"
fi
docker build -t $DOCKER_IMAGE $DIST_DIR $QUIET
echo -e "\e[96mDocker image build done.\e[0m"
echo
```

dist/sgx/scripts/setup
```diff
@@ -72,121 +54,49 @@ function checkForPinFile() {
...
-print_info "Installing the powHSM..."
-installPowHsm
-echo
-print_info "Starting the powHSM..."
-startPowHsm
-echo
 print_warning "Onboarding the powHSM..."
 onboard
-print_warning "Onboarding complete."
+print_info "Onboarding complete."
```
amendelzon, line 87: Why did we change `print_warning` with `print_info` and viceversa? Was that part of the improvements?

amendelzon (commented):
Some more questions

dist/sgx/scripts/setup
italo-sampaio: No strong reason, I just noticed the usage seemed a bit random as it was and tried to make it a bit more consistent:

warning / yellow -> I'm about to do something
info / green -> The current step succeeded
error / red -> The current step (or substep) failed

dist/sgx/scripts/run_with_docker
```diff
@@ -5,6 +5,13 @@ DIST_DIR=$(pwd)
 popd > /dev/null
 DOCKER_IMAGE=powhsmsgx:setup
 
+# HSM scripts directory
+SCRIPTS_DIR=$DIST_DIR/scripts
```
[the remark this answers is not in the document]
italo-sampaio: Oh nice catch. It was used for print_info / print_error before the latest changes, now it's unused

dist/sgx/scripts/setup
amendelzon: I understand now. I say let's leave it as is and treat that as a separate improvement since we'd also have to take a look at/change the way it's done for the Ledger scripts.

amendelzon: I thought we'd agreed on leaving the color scheme as it currently is on master?

italo-sampaio:
> I think we're still missing rolling back the color scheme improvements.

Oh, I think I misunderstood your previous comment on it. I thought we would leave it as it is in this commit. I'll add that change

italo-sampaio:
ready for review

amendelzon (approved):
LGTM

PR #344, opened 2025-04-15
SGX upgrade spec authentication

amendelzon:
- Implemented SGX upgrade spec authentication via provision of valid signatures
- SGX build and distribution build now require specifying an upgrades signer file
- Factored out Ethereum constants into `eth.h`
- Factored out Ledger firmware upgrade signer authorizers inclusion logic into `upgrade_signers.h`
- Moved existing firmware upgrade authorizers to common location
- Added spec authentication unit test cases to upgrade module unit tests
- Fixed failing unit tests


italo-sampaio (approved):
LGTM!

PR #350, opened 2025-05-05
Add SGX migration authorization tool

italo-sampaio:
Introduces a new tool for creating and signing SGX migration authorizations between enclaves. The tool follows a similar pattern to the existing signapp.py tool.

### DIFF
--- a/middleware/signmigration.py
+++ b/middleware/signmigration.py
@@ -1,0 +1,246 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2021 RSK Labs Ltd
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
+# use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+# of the Software, and to permit persons to whom the Software is furnished to do
+# so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import sys
+from os.path import isfile
+from argparse import ArgumentParser
+import ecdsa
+import logging
+from admin.misc import (
+    get_eth_dongle,
+    dispose_eth_dongle,
+    info,
+    AdminError
+)
+from comm.utils import is_hex_string_of_length
+from comm.bip32 import BIP32Path
+from admin.sgx_migration_authorization import SGXMigrationAuthorization, SGXMigrationSpec
+from admin.ledger_utils import eth_message_to_printable
+
+# Default signing path
+DEFAULT_ETH_PATH = "m/44'/60'/0'/0/0"
+
+
+def _require_output_path(options, require_existing=False):
+    if options.output_path is None:
+        raise AdminError("Must provide an output path (-o/--output)")
+    if require_existing and not isfile(options.output_path):
+        raise AdminError(f"Invalid output path: {options.output_path}")
+
+
+def do_message(options):
+    if options.exporter_hash is None:
+        raise AdminError("Must provide an exporter hash (-e/--exporter)")
+    if options.importer_hash is None:
+        raise AdminError("Must provide an importer hash (-i/--importer)")
+
+    info("Computing the SGX migration authorization message...")
+    migration_spec = SGXMigrationSpec({
+        "exporter": options.exporter_hash,
+        "importer": options.importer_hash
+    })
+    sgx_authorization = SGXMigrationAuthorization.for_spec(migration_spec)
+    if options.output_path is None:
+        info(eth_message_to_printable(migration_spec.get_authorization_msg()))
+    else:
+        sgx_authorization.save_to_jsonfile(options.output_path)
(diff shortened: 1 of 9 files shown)


amendelzon:
Could we please mark this as draft while it's not ready? Thanks

middleware/signmigration.py
```diff
@@ -0,0 +1,246 @@
...
+            info(f"Opening public key file {options.output_path}...")
+            info("Adding public key...")
+            with open(options.output_path, "w") as file:
+                file.write("%s\n" % pubkey.hex())
+            info(f"Public key saved to {options.output_path}")
+            return
+
+        # Is there an existing migration authorization? Read it
+        sgx_authorization = None
+        if not isfile(options.output_path):
```
amendelzon, line 133: Don't we have a function to check this?

middleware/signmigration.py
```diff
@@ -0,0 +1,246 @@
...
+    except AdminError:
+        raise
+    except Exception as e:
+        raise AdminError(f"Error signing with dongle: {e}")
+    finally:
+        dispose_eth_dongle(eth)
+
+
+def main():
+    logging.disable(logging.CRITICAL)
```
amendelzon, line 165: Why are we doing this?

amendelzon (commented):
Looks good, left a couple of questions

middleware/signmigration.py
italo-sampaio: Nice catch, yes, I'll replace it with the function.

middleware/signmigration.py
italo-sampaio: TBH, I just copied this from [signapp.py](https://github.com/rsksmart/rsk-powhsm/blob/master/middleware/signapp.py#L50).

Now that you mentioned, this apparently has no effect, since we're logging using `misc.info`. Should this be removed (probably from signapp.py as well)?

amendelzon: sounds good. could you please make sure the output and behaviour remain unaltered after the changes? thanks!

middleware/tests/admin/test_sgx_authorization.py
```diff
@@ -0,0 +1,333 @@
...
+    def test_add_signature(self):
+        # Test adding a valid signature
+        new_sig = "3045022100d2dac5b641d6a454cacdff045ab428bfc4c86e"\
+                  "004ff69728050a33788f6e9e7602207e8b3536a7a50185e2"\
+                  "7219237358823b98678fe32aa7a30b31155cbffad3747d"
+        self.sa.add_signature(new_sig)
+        self.assertEqual(self.sa.signatures, self.sigs + [new_sig])
+
+    def test_add_duplicate_signature(self):
+        # Adding the same signature should be allowed
```
amendelzon, line 101: Why is this allowed?

middleware/tests/admin/test_sgx_authorization.py
```diff
@@ -0,0 +1,333 @@
...
+    def test_authorization_message(self):
+        # Test authorization message format
+        expected_msg = (b"\x19Ethereum Signed Message:\n" +
+                        b"160" +
+                        b"RSK_powHSM_SGX_upgrade_from_" +
+                        self.exporter_mrenclave.encode("ASCII") +
+                        b"_to_" + self.importer_mrenclave.encode("ASCII"))
+        self.assertEqual(expected_msg, self.sa.migration_spec.get_authorization_msg())
+
+    def test_authorization_digest(self):
```
amendelzon, line 262: Isn't this a migration spec unit test?

middleware/tests/admin/test_sgx_authorization.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def test_to_dict(self):
+        # Test dictionary conversion
+        expected_dict = {
+            "exporter": self.exporter_mrenclave,
+            "importer": self.importer_mrenclave,
+        }
+        self.assertEqual(expected_dict, self.migration_spec.to_dict())
+
+    def test_authorization_message(self):
```
amendelzon, line 320: Should we also add an additional non-ethereum prefixed message generation test case?

middleware/tests/admin/test_sgx_authorization.py
```diff
@@ -0,0 +1,333 @@
...
+        # Test authorization message format
+        expected_msg = (b"\x19Ethereum Signed Message:\n" +
+                        b"160" +
+                        b"RSK_powHSM_SGX_upgrade_from_" +
+                        self.exporter_mrenclave.encode("ASCII") +
+                        b"_to_" + self.importer_mrenclave.encode("ASCII"))
+        self.assertEqual(expected_msg, self.migration_spec.get_authorization_msg())
+
+    def test_authorization_digest(self):
```
amendelzon, line 329: I think this is the same as the one on the migration authorisation, I'd remove the other one

middleware/tests/admin/test_signmigration.py
amendelzon, line 1: We need some unhappy path test cases.

amendelzon (approved):
LGTM!

All the comments I left are improvements to the unit tests. Let's please get this merged in as-is and then tackle those comments as a separate PR. Thanks!

middleware/tests/admin/test_sgx_authorization.py
italo-sampaio: Nice catch, you're right, this is duplicated.

middleware/tests/admin/test_sgx_authorization.py
italo-sampaio: Sounds good, I'll add a test case for it.

middleware/tests/admin/test_sgx_authorization.py
italo-sampaio: Yes, you're right. I'll remove the other one and leave only this one

middleware/tests/admin/test_sgx_authorization.py
italo-sampaio: Good question, not sure if it should TBH. I just reproduced signapp's behavior.

We can either modify the logic to turn this into an error or just remove this test case if it's not a big deal, what do you think?

italo-sampaio: As aligned offline, this behavior will be changed so that we don't allow duplicated signatures. The test case will be updated accordingly.

middleware/tests/admin/test_sgx_authorization.py
italo-sampaio: This will be tackled in an upcoming PR

middleware/tests/admin/test_sgx_authorization.py
italo-sampaio: This will be tackled in an upcoming PR

middleware/tests/admin/test_signmigration.py
italo-sampaio: This will be tackled in an upcoming PR

PR #353, opened 2025-05-09
SGX upgrade shared key computation

amendelzon:
- Added private/public keypair generation and addition to self evidence
- Added peer public key extraction upon peer evidence parsing
- Added shared secret computation for data migration
- Removed dummy key
- Added custom claim finder function to evidence module
- Completed evidence function docs
- Added shared key computation unit tests
- Added `evidence_get_custom_claim` function unit tests
- Fixed failing unit tests


italo-sampaio (approved):
LGTM!

PR #355, opened 2025-05-12
SGX upgrade spec auth: replacing sha-256 with keccak-256

amendelzon:
SGX upgrade spec authentication: replacing sha-256 with keccak-256 for message digest

- Updated hash algorithm
- Fixed failing unit tests


italo-sampaio (approved):
LGTM!

PR #357, opened 2025-05-19
Uninstall script for SGX setup

italo-sampaio:
Adds a standalone bash script for completely removing the current SGX powHSM installation from the machine. This can be used before a new onboarding to ensure that all the previously installed assets are cleared.

### DIFF
--- a/dist/misc/uninstall-sgx-powhsm
+++ b/dist/misc/uninstall-sgx-powhsm
@@ -23,6 +23,13 @@ function promptUserConfirmation() {
     fi
 }
 
+function checkExistingBackup() {
+    if [ -f "backup.tgz" ]; then
+        echo "Error: backup.tgz already exists. Please remove it first."
+        exit 1
+    fi
+}
+
 function backupInstallDirectory() {
     if [ ! -f "/etc/systemd/system/$SERVICE_NAME" ]; then
         echo "Service unit file not found, skipping backup."
@@ -40,13 +47,8 @@ function backupInstallDirectory() {
         return
     fi
 
-    if [ -f "backup.tgz" ]; then
-        echo "Error: backup.tgz already exists. Please remove it first."
-        exit 1
-    fi
-
     echo "Creating backup of $WORKING_DIR..."
-    tar czf backup.tgz -C "$(dirname "$WORKING_DIR")" "$(basename "$WORKING_DIR")"
+    tar czf backup.tgz -C $WORKING_DIR $WORKING_DIR
     error "Failed to create backup."
 
     echo "Backup created successfully! Find the previous installation in the backup.tgz file."
@@ -119,6 +121,7 @@ echo "Welcome to the SGX powHSM uninstall utility"
 echo
 
 promptUserConfirmation
+checkExistingBackup
 echo
 removeService
 echo
--- a/dist/misc/uninstall-sgx-powhsm
+++ b/dist/misc/uninstall-sgx-powhsm
@@ -1,0 +1,130 @@
+#!/bin/bash
+
+# The name of the service unit that will be uninstalled
+SERVICE_NAME=powhsmsgx.service
+# The internal network name for the powHSM service
+NETWORK=powhsmsgx_net
+# The user name created during the installation
+USER=powhsm
+
+function error() {
+    if [ $? -ne 0 ]; then
+        echo "Error: $1"
+        exit 1
+    fi
+}
+
+function promptUserConfirmation() {
+    echo "Are you sure you want to uninstall the powHSM SGX service from this machine?"
+    read -p "Proceed? [y/N]: " confirm
+    if [ "$confirm" != "Y" ] && [ "$confirm" != "y" ]; then
+        echo "Uninstall cancelled."
+        exit 1
+    fi
+}
+
+function backupInstallDirectory() {
+    if [ ! -f "/etc/systemd/system/$SERVICE_NAME" ]; then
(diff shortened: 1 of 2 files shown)


dist/misc/uninstall-sgx-powhsm
```diff
@@ -0,0 +1,130 @@
...
+        return
+    fi
+
+    if [ -f "backup.tgz" ]; then
+        echo "Error: backup.tgz already exists. Please remove it first."
+        exit 1
+    fi
+
+    echo "Creating backup of $WORKING_DIR..."
+    tar czf backup.tgz -C "$(dirname "$WORKING_DIR")" "$(basename "$WORKING_DIR")"
```
amendelzon, line 49: Why the `basename` portion? Doesn't that remove any absolute pathing that we'd probably want to keep?

amendelzon: And also why the `dirname` portion? Doesn't simply using `tar czf backup.tgz -C $WORKING_DIR $WORKING_DIR` work as expected? Assuming $WORKING_DIR is an absolute path (otherwise we could use `WORKING_DIR=$(realpath $WORKING_DIR)` before actually attempting to create the archive?).

dist/misc/uninstall-sgx-powhsm
```diff
@@ -0,0 +1,130 @@
...
+        echo "Could not determine working directory from service unit. Aborting."
+        exit 1
+    fi
+
+    if [ ! -d "$WORKING_DIR" ]; then
+        echo "Working directory $WORKING_DIR does not exist, skipping backup."
+        return
+    fi
+
+    if [ -f "backup.tgz" ]; then
```
amendelzon, line 43: Can we check this at the start, before actually doing anything? Otherwise e.g. the service would be disabled but then uninstallation wouldn't proceed.

amendelzon (changes requested):
I think we might need to do a bit of work on this one.

dist/misc/uninstall-sgx-powhsm
italo-sampaio: I didn't think keeping the entire absolute path was the desired output. I will change it as you suggested, thanks

amendelzon: It is not the desired output. The idea is for the archive to be from the perspective of within the installation directory. Meaning we don't want to keep the installation directory name, which can change depending on the installation and is uninteresting. I understand the current approach keeps it, right?

dist/misc/uninstall-sgx-powhsm
```diff
@@ -0,0 +1,130 @@
...
+    echo "Removing $SERVICE_NAME..."
+    rm -f /etc/systemd/system/$SERVICE_NAME
+    error "Could not remove $SERVICE_NAME."
+
+    echo "Reloading systemd daemon..."
+    systemctl daemon-reload
+    error "Could not reload systemd daemon."
+
+    echo "Resetting failed state for $SERVICE_NAME (if any)..."
+    systemctl reset-failed "$SERVICE_NAME" &> /dev/null
```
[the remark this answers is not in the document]
amendelzon: I believe there's an additional file created alongside the copied original service file. Maybe try removing that one too? Or is that not the problem.

dist/misc/uninstall-sgx-powhsm
italo-sampaio: > Meaning we don't want to keep the installation directory name, which can change depending on the installation and is uninteresting.

In this case, I believe what we want is `tar czf backup.tgz -C $WORKING_DIR .`, isn't it? In this case we package everything inside `WORKING_DIR` without any parent directories (including the directory name)

dist/misc/uninstall-sgx-powhsm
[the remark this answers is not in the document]
italo-sampaio: It's not the problem. The problem is that if the service was in a failed state before removal, it will still be listed on `systemctl list-units --full --all` even after stopped, disabled, etc. It will be listed as something like `powhsmsgx.service loaded failed failed ...`. This command is specifically to clean this state.

The installer script uses `systemctl list-units --full --all` to check if there are leftovers from a previous installation and aborts if it finds any, so it's probably a good idea to clear this flag after uninstalling.

dist/misc/uninstall-sgx-powhsm
amendelzon: Yes, thank you.

dist/misc/uninstall-sgx-powhsm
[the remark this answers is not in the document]
amendelzon: ok, got it. thanks.

amendelzon (changes requested):
Something's off with the commits. Maybe something with the rebase?

Also: are we not removing the installation directory at all? I didn't find that portion.

dist/misc/uninstall-sgx-powhsm
```diff
@@ -0,0 +1,133 @@
...
+if ! [ "$(id -u)" == "0" ]; then
+    echo "Please run with sudo."
+    exit 1
+fi
+
+echo "Welcome to the SGX powHSM uninstall utility"
+echo
+
+promptUserConfirmation
+checkExistingBackup
```
amendelzon, line 124: This could be checked before asking for user confirmation, right?

amendelzon (changes requested):
One additional comment

amendelzon (approved):
LGTM

PR #360, opened 2025-05-23
Ensures canonical DER encoding for signmigration tool

italo-sampaio:
- Updates signature generation to enforce canonical DER encoding
- Adds unit tests

### DIFF
--- a/middleware/tests/admin/test_signmigration.py
+++ b/middleware/tests/admin/test_signmigration.py
@@ -358,6 +358,38 @@ class TestSignMigrationKey(TestCase):
         )
         migration_auth_mock.from_jsonfile.assert_not_called()
 
+    def test_canonical_signature_encoding(self, _, migration_auth_mock, isfile_mock):
+        migration_auth = Mock()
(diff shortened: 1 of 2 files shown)


middleware/tests/admin/test_signmigration.py
```diff
@@ -358,6 +358,38 @@ def test_non_existent_output_file(self, info_mock, migration_auth_mock, isfile_m
         )
         migration_auth_mock.from_jsonfile.assert_not_called()
 
+    def test_canonical_signature_encoding(self, _, migration_auth_mock, isfile_mock):
+        migration_auth = Mock()
+        migration_auth_mock.from_jsonfile.return_value = migration_auth
+        migration_auth.add_signature.return_value = None
+        isfile_mock.return_value = True
+
+        # Use a digest that could produce non-canonical signature
```
amendelzon, line 367: This doesn't actually guarantee much. There are canonical signatures produced with these kind of keys. I think the best way is mocking the `sigencode_der_canonize` method just to make sure that is exactly what is called. Can we do that?

PR #369, opened 2025-06-09
Removes unnecessary port 777 exposure from SGX powHSM Service

italo-sampaio:
- Changes the service start script so that the sgx container is only accessible within the internal docker network.
- Incidentally fixes verbose parameter handling in start script


amendelzon (commented):
Looks good. Before approving: does this work in our current default powpeg setup?

italo-sampaio:
> Looks good. Before approving: does this work in our current default powpeg setup?

I tested locally and it's working, but will trigger a fully automated build and perform a full-cycle test just to be sure. I'll get back to you to confirm.

amendelzon (approved):
Thanks for testing, LGTM

PR #379, opened 2025-06-24
Fixes firmware and middleware coverage badges

italo-sampaio:
Fixes URL for uploading/displaying firmware and middleware reports and badge to/from powhsm_head bucket instead of powhsm_5.5.x


amendelzon (approved):
LGTM

PR #381, opened 2025-06-27
SGX blockchain operations performance improvement

amendelzon:
- Added a new meta-blockchain operation chunk-breakdown proxy for advance blockchain and update ancestor operations
- Overriding advance blockchain and update ancestor operations in SGX, replacing them with the meta-blockchain proxy
- Updating existing powHSM code to account for a potentially bigger-than-256-bytes APDU buffer
- Fixed minor emerging bugs appearing when the APDU buffer is bigger than 256 bytes
- Added support for resetting HSM operations from outside the hsm module
- HSM external processor modules are now responsible for resetting the HSM operations
- Added support for setting the communication buffer in the HAL
- Updated SGX middleware behavior for chunk-based operations
- Fixed failing firmware and middleware unit tests
- Added system unit tests for meta_bc overrides
- Added meta_bc unit tests
- Added SGX chunk-related middleware operations unit tests
- Factored out hsm2dongle unit tests for large operations (advance blockchain, update ancestor, authorised signing)
- Separated firmware test cases for SGX and non-SGX APDU sizes
- Incidentally disabling enclave logging by default, and using a new environment variable to control it (enabled for debug and simulator builds)
- Incidentally adding missing __init__.py to some unit test modules


italo-sampaio (approved):
LGTM!

PR #383, opened 2025-06-30
Add missing unit tests to hal/sgx/src/trusted

italo-sampaio:
- Added unit tests for access module
- Added unit tests for hash module


amendelzon (commented):
Looks good, left a comment regarding naming that would be good if we address.

amendelzon (approved):
LGTM

PR #400, opened 2025-08-18
Fix file ownership after upgrade

italo-sampaio:
Replicates the same strategy used in `dist/sgx/scripts/install_service`. The upgrade script is executed by `root` in the host machine, which results in leaving all files in the new INSTALL_DIR (including the directory itself) owned by that user. This means that the powhsm service does not have write permission to these files, since the service is executed by the `powhsm` user.

This is not a problem in the initial installation, since we explicitly change ownership in the `install_service` script. This commit replicates the same strategy to `upgrade-existing-powhsm`, ensuring that the new installation is owned by `powhsm` as well.


amendelzon (dismissed):
LGTM

amendelzon (changes requested):
LGTM, but we need to target ` develop-5.5`

PR #410, opened 2025-08-29
Failsafe verification of the custom claims hash when verifying evidence

amendelzon:
- Always checking the custom claims hash is contained within the evidence body in evidence_verify_and_extract_claims, regardless of whether claims are being extracted or not
- Added test cases for the new validations
- Added missing hash size constants to HAL hash module
- Fixed failing unit tests


italo-sampaio (approved):
LGTM!

PR #411, opened 2025-08-29
Adds missing unit tests for `hal/common_linked/src` modules

italo-sampaio:
Adds unit tests for the following modules:

- communication
- hash
- exceptions

Incidentally, adds a bugfix to the exception library that fixes the implementation of the `CATCH_L` macro.


amendelzon (changes requested):
Let's please also add the fix for the Ledger exceptions library. We should now be using [this new tag](https://github.com/rsksmart/rsk-powhsm-nanos-secure-sdk/releases/tag/nanos-1314-powhsm-v3) for the nanos repository.

PR #446, opened 2025-10-03
Add SGX static analysis to gen-static-analysis script and CI

italo-sampaio:
- Adds SGX static analysis to `gen-static-analysis` script
- Fixes all findings from SGX static analysis

Detailed explanations of each fix in the commit messages

--
### Failure on Ledger only

https://github.com/rsksmart/rsk-powhsm/actions/runs/18328597988

### Failure on SGX only

https://github.com/rsksmart/rsk-powhsm/actions/runs/18328673523

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -7,7 +7,7 @@ permissions: read-all
 
 jobs:
   static-analysis:
-    name: Run ledger static analysis
+    name: Run static analysis
     runs-on: ubuntu-latest
 
     steps:
@@ -17,9 +17,13 @@ jobs:
       - name: Build the ledger docker image
         run: docker/ledger/build
 
+      - name: Build the SGX docker image
+        run: docker/sgx/build
+
       - name: Run static analysis
         id: static-analysis
         run: |
+          docker/sgx/do-notty /hsm2/firmware/src/sgx "make generate-private-key"
           firmware/static-analysis/gen-static-analysis
       
       - name: Upload static analysis reports


firmware/src/common/src/bigdigits.c
amendelzon, line 1: Why are we making changes to this file?

docker/sgx/Dockerfile
```diff
@@ -6,6 +6,7 @@ RUN apt-get update && \
     apt-get install -y tar && \
     apt-get install -y curl && \
     apt-get install -y open-enclave=0.19.4 && \
+    apt-get install -y clang-tools && \
```
amendelzon, line 9: Does adding this affect at all the result of the normal SGX builds? Do the MRENCLAVE values, for example, get altered?

.github/workflows/static-analysis.yml
amendelzon, line 1: Have you double checked that an error in either the UI or the Signer fails the workflow but an error in SGX doesn't?

docker/sgx/Dockerfile
italo-sampaio: I don't see how it would affect the build, since we use `gcc`, but I'll double-check just in case.

italo-sampaio:
NOTE: see that the workflow name has changed from `Run ledger static analysis` to `Run static analysis`, so the rules for merging need to be updated as well

.github/workflows/static-analysis.yml
amendelzon: No need, thanks

amendelzon (approved):
LGTM

PR #447, opened 2025-10-07
Introduce 5.5.2 changes

amendelzon:
This includes all changes introduced from `5.5.1` to `5.5.2`, except for those that did not apply due to either being specifically for version `5.5.1` or due to already being present in `master`.

All 16 commits have been cherry picked with reference preservation, so that we can see the original commit and the original pull request.


italo-sampaio (approved):
LGTM!

PR #499, opened 2026-03-04
Fixed CRL type detection on get_intel_pcs_x509_crl

amendelzon:
- Fixed CRL type detection on `get_intel_pcs_x509_crl`
- Bubbling up errors for early detection


italo-sampaio (approved):
LGTM!

PR #500, opened 2026-03-17
Authorized signing MP buffer size and node count validations

amendelzon:
Added firmware test cases and validations for:
- Authorized signing MP chunk being longer than the sign authorizer expects
- Authorized signing MP with zero node count


italo-sampaio (approved):
LGTM!

PR #503, opened 2026-03-19
Fixed integer underflow bug in trie parser

amendelzon:
- Disallowed trie nodes that signal a shared prefix with a varint length and specify length zero
- Added unit test case


italo-sampaio (approved):
LGTM!

PR #505, opened 2026-04-13
Reduce manager log verbosity at INFO level

italo-sampaio:
Emit per-command INFO summaries from the ledger layer
Prior to this commit, the manager would emit each request twice at INFO:
once as raw wire JSON in the server layer, and again as a parsed-dict dump
in the protocol layer. Large commands such as advanceBlockchain produce
tens of KB of hex per call on both sides, which dominates the daily log volume.

This commit trims the redundancy while preserving operator visibility:

- Server layer: unchanged from master. The raw wire JSON remains at INFO
  so operators still see the literal bytes of every request and response.
- Comm protocol layer: the generic "In %s" / "Out %s" dumps are demoted
  from INFO to DEBUG.
- Ledger protocol layer: each command handler now emits a single
  human-readable INFO line at entry, describing what is being attempted
  in domain terms. Examples:
    - "Advance blockchain with 31 blocks"
    - "Sign authorized tx request for key id 'm/44'/1'/0'/0/0' (input 0)"
    - "Sign unauthorized hash request for key id 'm/44'/137'/0'/0/0'"
    - "Update ancestor with 21 blocks"


middleware/ledger/protocol.py
```diff
@@ -268,6 +269,9 @@ def _sign(self, request):
             if message_validation < self.ERROR_CODE_OK:
                 return (message_validation,)
 
+            self.logger.info(
+                "Sign unauthorized hash request for key id '%s'", request["keyId"]
```
amendelzon: We need to also show the hash here.

middleware/ledger/protocol.py
```diff
@@ -296,6 +300,10 @@ def _sign(self, request):
             # Shorthand
             msg = request["message"]
 
+            self.logger.info(
+                "Sign authorized tx request for key id '%s' (input %d)",
```
amendelzon: We need to show the (unsigned) BTC tx hash here as well.

middleware/ledger/protocol.py
```diff
@@ -463,6 +480,7 @@ def _get_blockchain_parameters(self, request):
             return (self.ERROR_CODE_DEVICE,)
 
     def _signer_heartbeat(self, request):
+        self.logger.info("Signer heartbeat")
```
amendelzon: We need to show the `udValue` here too.

middleware/ledger/protocol.py
```diff
@@ -491,6 +509,7 @@ def _signer_heartbeat(self, request):
             return (self.ERROR_CODE_DEVICE,)
 
     def _ui_heartbeat(self, request):
+        self.logger.info("UI heartbeat")
```
amendelzon: Same here regarding `udValue`.

middleware/tests/ledger/test_protocol.py
amendelzon, line 1: I generally don't think we need to test logging output, so feel free to remove these tests.

amendelzon (changes requested):
LGTM, a few things we need to add and we'll be good to go.

italo-sampaio:
Addressed all suggestions, ready for review again @amendelzon

middleware/comm/bitcoin.py
```diff
@@ -23,6 +23,36 @@
 import bitcoin.core
 
 
+def get_tx_hash_for_unsigned_tx(raw_tx_hex):
```
amendelzon: We don't need to "unsign" the tx in this case, just get the hash of the transaction as-is.

middleware/ledger/protocol.py
```diff
@@ -296,6 +302,18 @@ def _sign(self, request):
             # Shorthand
             msg = request["message"]
 
+            # Compute the unsigned tx hash for the log line
+            try:
+                tx_hash = get_tx_hash_for_unsigned_tx(msg["tx"])
```
amendelzon: We don't want to fail the operation if don't manage to compute the tx hash, since this is only for logging purposes. Instead, just log something reflecting the fact that it was impossible to compute the tx hash.

middleware/tests/comm/test_bitcoin.py
amendelzon, line 1: I would drop all the partially signed/unsigned cases, since we don't support legacy txs anymore. Just pick a couple of mainnet BTC transactions with their hashes and make sure the function computes the hash properly.

amendelzon (changes requested):
Almost good to go. A few things we need to tweak a bit further.

amendelzon (approved):
LGTM!

PR #507, opened 2026-04-20
Pin mware apt sources to snapshot.debian.org

italo-sampaio:
Switches docker/mware/Dockerfile from upstream Debian repos to the snapshot.debian.org archive at a fixed date. This stops the recurring build failures caused by pinned apt versions being rotated out of the main Debian repos, while preserving the version pins.

### DIFF
--- a/docker/mware/Dockerfile
+++ b/docker/mware/Dockerfile
@@ -1,8 +1,14 @@
 FROM python:3.12-slim-bookworm@sha256:10f3aaab98db50cba827d3b33a91f39dc9ec2d02ca9b85cbc5008220d07b17f3
 
 ARG PIP_TIMEOUT=60
+ARG SNAPSHOT_DATE=20260415T000000Z
 
 WORKDIR /hsm2
+
+RUN rm -f /etc/apt/sources.list.d/debian.sources && \
+    echo "deb [check-valid-until=no] https://snapshot.debian.org/archive/debian/${SNAPSHOT_DATE}/ bookworm main" > /etc/apt/sources.list && \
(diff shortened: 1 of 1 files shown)


docker/mware/Dockerfile
```diff
@@ -1,9 +1,15 @@
 FROM python:3.12-slim-bookworm@sha256:10f3aaab98db50cba827d3b33a91f39dc9ec2d02ca9b85cbc5008220d07b17f3
 
 ARG PIP_TIMEOUT=60
+ARG SNAPSHOT_DATE=20260415T000000Z
 
 WORKDIR /hsm2
 
+RUN rm -f /etc/apt/sources.list.d/debian.sources && \
+    echo "deb [check-valid-until=no] https://snapshot.debian.org/archive/debian/${SNAPSHOT_DATE}/ bookworm main" > /etc/apt/sources.list && \
```
amendelzon, line 9: I'd like to get @VaWheel 's opinion on the (necessary) use of `[check-valid-until=no]` here.

amendelzon (commented):
We'll wait for @VaWheel 's feedback before moving forwards with this.

docker/mware/Dockerfile
amendelzon: According to @VaWheel :
> Yes this is ok for this case. We already pin dependencies to concrete versions, with snapshot and [check-valid-until=no] we'll pin everything else that is not explicitly pinned. The main risk is that some old vulnerable package remain unfixed. But this is only a build environment and not deployed runtime, so risk is acceptable here

So we should be good to go.

amendelzon (approved):
LGTM

PR #514, opened 2026-06-24
Added maximum receipt depth validation

amendelzon:
- Preventing receipt parsing from going past the maximum allowed receipt depth
- Covering the case with a new test


italo-sampaio (approved):
LGTM!

